PR #4, opened 2018-05-07
Revert the picmg discover messages back to DEBUG.

tabraham:
commit 58d510f90feb ("ID : 103 picmg discover messages should be DEBUG, 
not INFO") changed picmg discover messages from DEBUG to INFO. However, 
commit f1c6118c722b ("ID:320 - Add VITA 46.11 support") reverted this without 
explanation or relevance.

This patch reverts the picmg discover messages back to DEBUG.

### DIFF
--- a/lib/ipmi_picmg.c
+++ b/lib/ipmi_picmg.c
@@ -2348,13 +2348,13 @@ picmg_discover(struct ipmi_intf *intf) {
 	req.msg.data_len = 1;
 	msg_data = 0;
 
-	lprintf(LOG_INFO, "Running Get PICMG Properties my_addr %#x, transit %#x, target %#x",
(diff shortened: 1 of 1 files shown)


AlexanderAmelkin:
I have a feeling that commit f1c6118 reverted more than just log level changes. Could you please also check if the rest of their changes to [ipmi_main.c](/ipmitool/ipmitool/commit/f1c6118c722ba4f20f3bdb2324503821a33b4e3d#diff-6be191d206bcd39125906410048b984f) aren't destructive? I don't have any PICMG or VITA hardware available.

tabraham:
The messages were the only issue reported by a customer using pcimg, so I would assume it's functioning as expected. Unfortunately, I don't have access to VITA hardware.

AlexanderAmelkin:
The change has been merged.

PR #16, opened 2018-06-18
mc: Always return negative value on error

n3world:
ipmitool main() only considers negative return values to be an error.
ipmi_sysinfo_main can return a positive error when a command fails. When
this occurs the mc command prints an error message but the return code
from ipmitool is 0. By negating the return code in ipmi_sysinfo_main
main will return a non 0 exit status.

Signed-off-by: Nate Clark <nate.clark@hitachivantara.com>

### DIFF
--- a/lib/ipmi_mc.c
+++ b/lib/ipmi_mc.c
@@ -1367,12 +1367,16 @@ ipmi_sysinfo_main(struct ipmi_intf *intf, int argc, char ** argv, int is_set)
 	if (rc < 0) {
 		lprintf(LOG_ERR, "%s %s set %d command failed", argv[0], argv[1], set);
 	}
-	else if (rc == 0x80) {
-		lprintf(LOG_ERR, "%s %s parameter not supported", argv[0], argv[1]);
-	}
 	else if (rc > 0) {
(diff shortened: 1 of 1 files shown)


AlexanderAmelkin:
I'm actually thinking about changing `exit(EXIT_SUCCESS)` to `exit(rc)` in `main()`.
That way only successful commands would return EXIT_SUCCESS (which is zero on all systems I know), and for all failed commands `ipmitool` would return a non-zero value that could be then treated as needed by the caller.

I'd like to reserve negative return values for `int` functions for "local" errors like argument errors or network errors. Actually, considering that only a single unsigned 8-bit value can be returned by `wait()` syscall, I want the only negative value (-1 => 255) reserved for all kinds of "local" errors while all other values I want to indicate IPMI completion codes.

Hence I believe this pull request need not to be merged. If you don't have any reasonable objections I'm going to reject it.

n3world:
That makes sense to change all of the return values from the different commands to be consistent. I would have preferred to do something like that but didn't want the change to be too invasive, since that would have required auditing all other commands and changing their return codes.

PR #32, opened 2018-08-13
Auto selection of 'best' cipher suite available

vmauery:
Based on current crypto alogrithms, one could rank cipher suites along
these lines:

17 > 3 >> all the rest

17 and 3 are the only cipher suites that implement any sort of
confidentiality alogorithm that is secure. In addition, any hmac-md5 or
md5 integrity algorithm used in integrity is weak at best and dangerous
for authentication.

This could possibly be enabled in a simpler mechanism by simply checking
for 17 and then choosing it before falling back to 3, but the way this
is implemented, it makes it easy to change the list of acceptable
algorithms from two to three or more items.

### DIFF
--- a/lib/ipmi_channel.c
+++ b/lib/ipmi_channel.c
@@ -342,75 +342,108 @@ ipmi_get_channel_auth_cap(struct ipmi_intf *intf, uint8_t channel, uint8_t priv)
 	return 0;
 }
 
+static inline size_t parse_cipher_suite(uint8_t *cipher_suite_data,
+                                        size_t data_len,
+                                        uint32_t *iana,
+                                        uint8_t *auth_alg,
+                                        uint8_t *integrity_alg,
+                                        uint8_t *crypt_alg,
+                                        enum cipher_suite_ids *cipher_suite_id)
+{
+	size_t size = 0;
+	const char *incomplete = "Incomplete data record in cipher suite data";
+
+	if (*cipher_suite_data == STANDARD_CIPHER_SUITE) {
+		struct std_cipher_suite_record_t *record =
+			(struct std_cipher_suite_record_t*)cipher_suite_data;
+
+		/* Verify that we have at least a full record left; id + 3 algs */
+		if (data_len < sizeof(*record)) {
+			lprintf(LOG_INFO, "%s", incomplete);
+			goto out;
+		}
+
+		/* IANA code remains default (0) */
+		*cipher_suite_id = record->cipher_suite_id;
+		*auth_alg = CIPHER_ALG_MASK & record->auth_alg;
+		*integrity_alg = CIPHER_ALG_MASK & record->integrity_alg;
+		*crypt_alg = CIPHER_ALG_MASK & record->crypt_alg;
+		size = sizeof(*record);
+	} else if (*cipher_suite_data == OEM_CIPHER_SUITE) {
+		/* OEM record type */
+		struct oem_cipher_suite_record_t *record =
+			(struct oem_cipher_suite_record_t*)cipher_suite_data;
+		/* Verify that we have at least a full record left
+		 * id + iana + 3 algs
+		 */
+		if (data_len < sizeof(*record)) {
+			lprintf(LOG_INFO, "%s", incomplete);
+			goto out;
+		}
+
+		/* Grab the IANA */
+		*iana = ipmi24toh(record->iana);
+		*cipher_suite_id = record->cipher_suite_id;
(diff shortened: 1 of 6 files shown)


src/plugins/lanplus/lanplus.c
```diff
@@ -3381,6 +3381,56 @@ ipmi_set_session_privlvl_cmd(struct ipmi_intf * intf)
...
+
+	if (_ipmi_get_channel_cipher_suites(intf, "ipmi", IPMI_LAN_CHANNEL_E,
+				suites, &nr_suites) < 0) {
+		return IPMI_LANPLUS_CIPHER_SUITE_3;
+	}
+	for (ipref = 0; ipref < nr_preferred &&
+			IPMI_LANPLUS_CIPHER_SUITE_RESERVED != best_suite; ipref++) {
+		for (i = 0; i < nr_suites; i++) {
+			if (cipher_order_preferred[ipref] ==
+					suites[i].cipher_suite_id) {
```
AlexanderAmelkin: And here too. Alignment and curly.

src/plugins/lanplus/lanplus.c
```diff
@@ -3454,6 +3504,15 @@ ipmi_lanplus_open(struct ipmi_intf * intf)
 		lprintf(LOG_INFO, "This BMC does not support IPMI v2 / RMCP+");
 		goto fail;
 	}
+	/*
+	 * If no cipher suite was provided, query the channel cipher suite list and
+	 * pick the best one available
+	 */
+	if (IPMI_LANPLUS_CIPHER_SUITE_RESERVED ==
+			intf->ssn_params.cipher_suite_id) {
```
AlexanderAmelkin: alignment and curly

src/plugins/lanplus/lanplus.h
```diff
@@ -78,6 +78,25 @@
...
+#define IPMI_LANPLUS_CIPHER_SUITE_8 8
+#define IPMI_LANPLUS_CIPHER_SUITE_9 9
+#define IPMI_LANPLUS_CIPHER_SUITE_10 10
+#define IPMI_LANPLUS_CIPHER_SUITE_11 11
+#define IPMI_LANPLUS_CIPHER_SUITE_12 12
+#define IPMI_LANPLUS_CIPHER_SUITE_13 13
+#define IPMI_LANPLUS_CIPHER_SUITE_14 14
+#define IPMI_LANPLUS_CIPHER_SUITE_15 15
+#define IPMI_LANPLUS_CIPHER_SUITE_16 16
+#define IPMI_LANPLUS_CIPHER_SUITE_17 17
```
AlexanderAmelkin: make these an enum type instead, please. enums are symbols that can be seen/resolved during gdb session, unlike macros. change the associated variables/struct fields to that type too, please.

lib/ipmi_channel.c
```diff
@@ -370,147 +379,124 @@ ipmi_get_channel_cipher_suites(struct ipmi_intf *intf, const char *payload_type,
...
 
 			/* Grab the IANA */
 			iana =
-				cipher_suite_data[offset]            | 
-				(cipher_suite_data[offset + 1] << 8) | 
-				(cipher_suite_data[offset + 2] << 16);
-			offset += 3;
+				record->iana[0]        |
+				(record->iana[1] << 8) |
+				(record->iana[2] << 16);
```
AlexanderAmelkin: There is an `ipmi24toh()` for this now. Please use it.

src/plugins/lanplus/lanplus.c
```diff
@@ -3381,6 +3381,56 @@ ipmi_set_session_privlvl_cmd(struct ipmi_intf * intf)
...
+{
+	/* 256 cipher suites may seem ridiculous, but get channel cipher suites
+	 * can return a maximum of 512 entries:
+	 *   16 * 0x40 bytes / 2 bytes per entry
+	 * though with 5 byte entries, it is 204, but likely will
+	 * only require fewer than 20 (standard defined cipher suites) */
+	struct cipher_suite_info suites[256];
+	size_t nr_suites = sizeof(suites)/sizeof(suites[0]);
+	/* cipher suite best order is chosen with this criteria:
+	 * hmac-md5 and md5 are BAD; xc4 is bad; aes128 is required
```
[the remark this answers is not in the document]
vmauery: Yes, it would be possible to just check for 17. But I wanted to make it so an arbitrary priority list could be substituted if other people had different opinions about which cipher suites were important to them.

vmauery:
I think I have addressed all the concerns you have raised.

PR #35, opened 2018-09-04
plugins: open: Properly enable event receiver

BenjaminFair:
The ioctl to enable the event receiver in the OpenIPMI interface is
called with an argument of 0, which tells OpenIPMI to disable it. Set
the argument to 1 instead so that it will be enabled.

Signed-off-by: Benjamin Fair <benjaminfair@google.com>

### DIFF
--- a/src/plugins/open/open.c
+++ b/src/plugins/open/open.c
@@ -91,8 +91,6 @@ extern int verbose;
 static int
 ipmi_openipmi_open(struct ipmi_intf * intf)
 {
-	int i = 1;
-
 	char ipmi_dev[16];
 	char ipmi_devfs[16];
 	char ipmi_devfs2[16];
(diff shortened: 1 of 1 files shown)


src/plugins/open/open.c
```diff
@@ -91,7 +91,7 @@ extern int verbose;
 static int
 ipmi_openipmi_open(struct ipmi_intf * intf)
 {
-	int i = 0;
+	int i = 1;
```
AlexanderAmelkin: Good catch. This ioctl is listed in manpages as 'Stub only', but apparently it not a stub anymore.
Nonetheless, I hate this `i` name for the variable. That is a nice name for an iterator, but is awful for an ioctl parameter. Since you're touching this line, it's a good moment to rename the variable to something meaningful like `receive_events`. Also, as `ipmitool` is now built according to C11, you can safely move the definition closer to the `ioctl()` using it.

I would also prefer `TRUE` used instead of `1`. That macro is available via `ipmitool/helper.h`.

BenjaminFair:
Done.

You said that you prefer comments to be addressed in a new commit rather than amending the existing one, right?

AlexanderAmelkin:
> You said that you prefer comments to be addressed in a new commit rather than amending the existing one, right?

Exactly! GitHub doesn't show old commit (unlike Bitbucket), so squashing them together just deletes them and makes it harder to review (especially in big pull requests). Thanks for reading the contribution guidelines.

PR #38, opened 2018-09-20
 ignore sol keepalive timeout when keepalives fail

kthommandra:
When a keepalive is requested prior to keepalive timeout,
the keepalive request is flagged as successful without sending
the request over wire. If BMC goes dark, after a SOL session
has been established, this fake success from keepalive request
can cause the ipmi_sol_red_pill to get stuck in a forever loop.
The fix is to not skip the keepalive requests after one of them
fails until it passes.

Testing:

ipmitool -I lanplus -H $host -U $user -P $passwd sol activate
iptables -A INPUT -s $host -j DROP
Without fix the ipmitool sol activate command just hangs with stack showing ipmi_sol_red_pill
With fix the ipmitool SOL session established in 1 should get dropped/closed after some time (retries of keepalive requests)

### DIFF
--- a/lib/ipmi_sol.c
+++ b/lib/ipmi_sol.c
@@ -1473,7 +1473,7 @@ processSolUserInput(
 }
 
 static int
-ipmi_sol_keepalive_using_sol(struct ipmi_intf * intf)
+ipmi_sol_keepalive_using_sol(struct ipmi_intf * intf, int force_keepalive)
 {
 	struct ipmi_v2_payload v2_payload;
 	struct timeval end;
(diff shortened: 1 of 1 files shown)


AlexanderAmelkin:
@vmauery, do you mind looking at this? I didn't read deep into this patch, but I'm not sure it's the correct way to fix the problem. I'll look into it later myself.

vmauery:
The current timeout works just fine on my system. I instrumented the code and printed out the keepAliveRet and retrySol values and followed the testing procedures above. It behaved just as expected: I saw 32 keepAliveRet = 0 for the faked retries and then retrySol=0...6 and then a timeout and closed session. This whole process takes about 1m40s because each dropped packet takes a while to timeout and then it retries 5 times.

I don't think this patch is necessary.

AlexanderAmelkin:
Just as @vmauery said, the current build of `ipmitool` without this change successfully times out in about  1:40 since putting the iptables drop rule in effect:

```
$ time src/ipmitool -H openbmc -P 0penBmc -I lanplus sol activate
[SOL Session operational.  Use ~? for help]
Error: No response to keepalive - Terminating session
Error: No response de-activating SOL payload

real    1m39.343s
user    0m0.004s
sys     0m0.004s
```
Thus, I don't think this change is actually needed. In fact I think this change will make SOL sessions much less robust on poor quality connections.

PR #39, opened 2018-09-22
ipmi_sensor: added csv output

eyJhb:
Added csv output for `ipmi_sensor.c`, which was much needed for a project of mine.
Should not interfere with the output, as they use dots for decimal places, and not commas.

### DIFF
--- a/lib/ipmi_sensor.c
+++ b/lib/ipmi_sensor.c
@@ -260,6 +260,184 @@ print_thresh_setting(struct sdr_record_full_sensor *full,
 	}
 }
 
+static void
+dump_sensor_fc_thredshold_csv(
+	int thresh_available,
+	const char *thresh_status,
+	struct ipmi_rs *rsp,
+	struct sensor_reading *sr)
+{
+	printf("%s", sr->s_id);
+	if (sr->s_reading_valid) {
+		if (sr->s_has_analog_value)
+			printf(",%.3f,%s,%s",
+			       sr->s_a_val, sr->s_a_units, thresh_status);
+		else
+			printf(",0x%x,%s,%s",
+			       sr->s_reading, sr->s_a_units, thresh_status);
+	} else {
+		printf(",%s,%s,%s",
+		       "na", sr->s_a_units, "na");
+	}
+	if (thresh_available && sr->full) {
+#define PTS(bit, dataidx) { \
+print_thresh_setting(sr->full, rsp->data[0] & (bit), \
+rsp->data[(dataidx)], ",", "%.3f", "0x%x", "%s"); \
+}
+		PTS(LOWER_NON_RECOV_SPECIFIED, 3);
+		PTS(LOWER_CRIT_SPECIFIED, 2);
+		PTS(LOWER_NON_CRIT_SPECIFIED, 1);
+		PTS(UPPER_NON_CRIT_SPECIFIED, 4);
+		PTS(UPPER_CRIT_SPECIFIED, 5);
+		PTS(UPPER_NON_RECOV_SPECIFIED, 6);
+#undef PTS
+	} else {
+		printf(",%s,%s,%s,%s,%s,%s",
+		       "na", "na", "na", "na", "na", "na");
+	}
+	printf("\n");
+}
+
+/* output format
+ *   id value units status thresholds....
+ */
+static void
+dump_sensor_fc_thredshold(
+	int thresh_available,
+	const char *thresh_status,
+	struct ipmi_rs *rsp,
+	struct sensor_reading *sr)
+{
+	printf("%-16s ", sr->s_id);
+	if (sr->s_reading_valid) {
+		if (sr->s_has_analog_value)
+			printf("| %-10.3f | %-10s | %-6s",
+			       sr->s_a_val, sr->s_a_units, thresh_status);
+		else
+			printf("| 0x%-8x | %-10s | %-6s",
+			       sr->s_reading, sr->s_a_units, thresh_status);
+	} else {
+		printf("| %-10s | %-10s | %-6s",
+		       "na", sr->s_a_units, "na");
+	}
+	if (thresh_available && sr->full) {
+#define PTS(bit, dataidx) { \
+print_thresh_setting(sr->full, rsp->data[0] & (bit), \
+rsp->data[(dataidx)], "| ", "%-10.3f", "0x%-8x", "%-10s"); \
+}
+		PTS(LOWER_NON_RECOV_SPECIFIED, 3);
+		PTS(LOWER_CRIT_SPECIFIED, 2);
+		PTS(LOWER_NON_CRIT_SPECIFIED, 1);
+		PTS(UPPER_NON_CRIT_SPECIFIED, 4);
+		PTS(UPPER_CRIT_SPECIFIED, 5);
+		PTS(UPPER_NON_RECOV_SPECIFIED, 6);
+#undef PTS
+	} else {
+		printf("| %-10s| %-10s| %-10s| %-10s| %-10s| %-10s",
+		       "na", "na", "na", "na", "na", "na");
+	}
+
+	printf("\n");
+}
+
+static void
+dump_sensor_fc_thredshold_verbose(
+	int thresh_available,
+	const char *thresh_status,
+	struct ipmi_intf *intf,
+	struct sdr_record_common_sensor *sensor,
+	struct ipmi_rs *rsp,
+	struct sensor_reading *sr)
+{
+	printf("Sensor ID              : %s (0x%x)\n",
+	       sr->s_id, sensor->keys.sensor_num);
+
+	printf(" Entity ID             : %d.%d\n",
+	       sensor->entity.id, sensor->entity.instance);
+
+	printf(" Sensor Type (Threshold)  : %s\n",
+	       ipmi_get_sensor_type(intf, sensor->sensor.type));
+
+	printf(" Sensor Reading        : ");
+	if (sr->s_reading_valid) {
+		if (sr->full) {
+			uint16_t raw_tol = __TO_TOL(sr->full->mtol);
+			if (sr->s_has_analog_value) {
+				double tol =
+					sdr_convert_sensor_tolerance(sr->full,
+								   raw_tol);
+				printf("%.*f (+/- %.*f) %s\n",
+				       (sr->s_a_val == (int)
+				       sr->s_a_val) ? 0 : 3,
+				       sr->s_a_val,
+				       (tol == (int) tol) ? 0 : 3, tol,
+				       sr->s_a_units);
+			} else {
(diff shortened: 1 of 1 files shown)


lib/ipmi_sensor.c
```diff
@@ -169,7 +169,25 @@ ipmi_sensor_print_fc_discrete(struct ipmi_intf *intf,
 	}
 
 	if (csv_output) {
-		/* NOT IMPLEMENTED */
+        printf("%s", sr->s_id);
```
AlexanderAmelkin: Please format according to project's [rules](https://github.com/ipmitool/ipmitool/wiki/Coding-Standards).

lib/ipmi_sensor.c
```diff
@@ -169,7 +169,25 @@ ipmi_sensor_print_fc_discrete(struct ipmi_intf *intf,
...
+        if (sr->s_reading_valid) {
+            if (sr->s_has_analog_value) {
+                /* don't show discrete component */
+                printf(",%s%s%s",
```
AlexanderAmelkin: Are you sure that, for instance, "0.97Vok"  is what you want to have in a cell? I believe there must be either spaces or field separators between those `%s` specifiers.

Anyway, this won't work unless you enclose values in double quotes as per [RFC4180](https://tools.ietf.org/html/rfc4180) because some locales use comma as a decimal point.

lib/ipmi_sensor.c
```diff
@@ -169,7 +169,25 @@ ipmi_sensor_print_fc_discrete(struct ipmi_intf *intf,
...
+                       sr->s_a_str, sr->s_a_units, "ok");
+            } else {
+                printf(",0x%x%s0x%02x%02x",
```
AlexanderAmelkin: "0x0discrete0x4000" ? Are you sure?

lib/ipmi_sensor.c
```diff
@@ -169,7 +169,25 @@ ipmi_sensor_print_fc_discrete(struct ipmi_intf *intf,
...
+                       sr->s_reading, "discrete",
+                       sr->s_data2, sr->s_data3);
+            }
+        } else {
+            printf(",%s%s%s",
+                   "na", "discrete", "na");
```
AlexanderAmelkin: "nadiscretena" ? What the heck?

lib/ipmi_sensor.c
```diff
@@ -270,7 +288,36 @@ ipmi_sensor_print_fc_threshold(struct ipmi_intf *intf,
...
+            else
+                printf(",0x%x,%s,%s",
+                       sr->s_reading, sr->s_a_units, thresh_status);
+        } else {
+            printf(",%s,%s,%s",
+                   "na", sr->s_a_units, "na");
+        }
+        if (thresh_available && sr->full) {
+#define PTS(bit, dataidx) {                     \
+print_thresh_setting(sr->full, rsp->data[0] & (bit),        \
```
AlexanderAmelkin: Indentation!
Continuation symbols (`\`) must be 1 space after the last character of code, not chaotically placed in the middle of nowhere.

lib/ipmi_sensor.c
```diff
@@ -270,7 +288,36 @@ ipmi_sensor_print_fc_threshold(struct ipmi_intf *intf,
...
+    rsp->data[(dataidx)], ",", "%.3f", "0xx", "%s");    \
+}
+            PTS(LOWER_NON_RECOV_SPECIFIED,  3);
+            PTS(LOWER_CRIT_SPECIFIED,   2);
```
AlexanderAmelkin: Why so many spaces after the comma? Why different amount of them on each line? Must be exactly 1 everywhere.

lib/ipmi_sensor.c
```diff
@@ -270,7 +288,36 @@ ipmi_sensor_print_fc_threshold(struct ipmi_intf *intf,
...
+                printf(",0x%x,%s,%s",
+                       sr->s_reading, sr->s_a_units, thresh_status);
+        } else {
+            printf(",%s,%s,%s",
+                   "na", sr->s_a_units, "na");
+        }
+        if (thresh_available && sr->full) {
+#define PTS(bit, dataidx) {                     \
+print_thresh_setting(sr->full, rsp->data[0] & (bit),        \
+    rsp->data[(dataidx)], ",", "%.3f", "0xx", "%s");    \
```
AlexanderAmelkin: `0xx` ?

lib/ipmi_sensor.c
```diff
@@ -336,16 +336,16 @@ print_thresh_setting(sr->full, rsp->data[0] & (bit),        \
 				       "na", sr->s_a_units, "na");
 			}
 			if (thresh_available && sr->full) {
-#define PTS(bit, dataidx) {						\
-	print_thresh_setting(sr->full, rsp->data[0] & (bit),  		\
-	    rsp->data[(dataidx)], "| ", "%-10.3f", "0x-8x", "%-10s");	\
+#define PTS(bit, dataidx) {	\
```
AlexanderAmelkin: Still doesn't look like 1 space

eyJhb:
Everything should be fixed now! What a mess! Really sorry to have put you through this... Not good enough from my side.

But hey, also fixed a small preexisting bug :+1:

lib/ipmi_sensor.c
```diff
@@ -288,36 +288,36 @@ ipmi_sensor_print_fc_threshold(struct ipmi_intf *intf,
...
-        } else {
-            printf
-                (",%s,%s,%s,%s,%s,%s",
-                 "na", "na", "na", "na", "na", "na");
-        }
-        printf("\n");
+		} else {
+			printf
+			     (",%s,%s,%s,%s,%s,%s",
+			     "na", "na", "na", "na", "na", "na");
```
AlexanderAmelkin: Please avoid breaking the line before the opening parenthesis. I know there are places in this very file where this has been done previously, but let's not introduce more of this.

I would appreciate if you broke this `ipmi_sensor_print_fc_threshold()` function with a separate commit into smaller pieces to make it shallower indented.

lib/ipmi_sensor.c
eyJhb: Fixed in latest commit 48fe375

lib/ipmi_sensor.c
eyJhb: First part has been fixed in 48fe375, but not quite sure about last part.
Firstly I am not the best C coder, but I could give a shot!

Do you mean something like this?
https://github.com/ipmitool/ipmitool/blob/c70665c3d1ae8d19634eb2ea386d1065a4cff506/lib/ipmi_user.c#L296-L302

AlexanderAmelkin: Yes, something like that.

eyJhb: How is 6223ec1? Is that good enough?

lib/ipmi_sensor.c
```diff
@@ -260,6 +260,183 @@ print_thresh_setting(struct sdr_record_full_sensor *full,
 	}
 }
 
+static void
+dump_sensor_fc_thredshold_csv(int thresh_available,
+			 const char *thresh_status,
```
AlexanderAmelkin: These lines are strangely aligned with a mix of tabs and spaces. Please use spaces only and align to `int` above.
Alternatively, put a line break after `(` and simply indent all arguments with a single TAB.

lib/ipmi_sensor.c
```diff
@@ -260,6 +260,183 @@ print_thresh_setting(struct sdr_record_full_sensor *full,
...
+			 struct ipmi_rs *rsp,
+			 struct sensor_reading *sr)
+{
+	printf("%s", sr->s_id);
+	if (sr->s_reading_valid) {
+		if (sr->s_has_analog_value)
+			printf(",%.3f,%s,%s",
+				   sr->s_a_val, sr->s_a_units, thresh_status);
```
AlexanderAmelkin: This is also strangely indented/aligned. Need 1 TAB less and spaces for alignment. Same applies to many lines below.

lib/ipmi_sensor.c
```diff
@@ -260,6 +260,183 @@ print_thresh_setting(struct sdr_record_full_sensor *full,
...
+{
+	printf("Sensor ID              : %s (0x%x)\n",
+		   sr->s_id, sensor->keys.sensor_num);
+
+	printf(" Entity ID             : %d.%d\n",
+		   sensor->entity.id, sensor->entity.instance);
+
+	printf(" Sensor Type (Threshold)  : %s\n",
+		   ipmi_get_sensor_type(intf, sensor->sensor.
+						 type));
```
AlexanderAmelkin: No need for a line break here

lib/ipmi_sensor.c
```diff
@@ -288,160 +465,12 @@ ipmi_sensor_print_fc_threshold(struct ipmi_intf *intf,
...
-			PTS(UPPER_NON_CRIT_SPECIFIED, 4);
-			PTS(UPPER_CRIT_SPECIFIED, 5);
-			PTS(UPPER_NON_RECOV_SPECIFIED, 6);
-#undef PTS
-		} else {
-			printf(",%s,%s,%s,%s,%s,%s",
-			       "na", "na", "na", "na", "na", "na");
-		}
-		printf("\n");
+        dump_sensor_fc_thredshold_csv(thresh_available, thresh_status, rsp, sr);
```
AlexanderAmelkin: Bad indentation.

lib/ipmi_sensor.c
```diff
@@ -288,160 +465,12 @@ ipmi_sensor_print_fc_threshold(struct ipmi_intf *intf,
...
-				PTS(UPPER_CRIT_SPECIFIED, 5);
-				PTS(UPPER_NON_RECOV_SPECIFIED, 6);
-#undef PTS
-			} else {
-				printf("| %-10s| %-10s| %-10s| %-10s| %-10s| %-10s",
-				       "na", "na", "na", "na", "na", "na");
-			}
-
-			printf("\n");
+			dump_sensor_fc_thredshold(thresh_available, thresh_status, rsp, sr);
```
AlexanderAmelkin: please break the line to fit in 80 characters.

lib/ipmi_sensor.c
```diff
@@ -288,160 +465,12 @@ ipmi_sensor_print_fc_threshold(struct ipmi_intf *intf,
...
-							   sensor_num,
-							   sensor->sensor.type,
-							   sensor->event_type,
-							   ANALOG_SENSOR,
-							   sensor->keys.owner_id,
-							   sensor->keys.lun,
-							   sensor->keys.channel);
-
-			printf("\n");
+			dump_sensor_fc_thredshold_verbose(thresh_available, thresh_status, intf, sensor, rsp, sr);
```
AlexanderAmelkin: please break the line to fit in 80 characters.

PR #45, opened 2018-10-29
exchange-bmc-os-info: Remove dependency on ipmi.service

charles-rose:
Most modern systems do not require ipmi.service to load the kernel
modules. Checking for /dev/ipmi* would be sufficient.

Signed-off-by: Charles Rose <charles.rose@dell.com>

### DIFF
--- a/contrib/exchange-bmc-os-info.service.redhat
+++ b/contrib/exchange-bmc-os-info.service.redhat
@@ -1,7 +1,7 @@
 [Unit]
 Description=Exchange Information between BMC and OS
-After=ipmi.service network.target
-Requires=ipmi.service
+ConditionFileIsExecutable=/usr/bin/ipmitool
+ConditionPathExistsGlob=/dev/ipmi*
 
 [Service]
 Type=oneshot


AlexanderAmelkin:
As far as I understand, the current implementation only runs this `exchange-bmc-os-info` unit after the ipmi drivers are known to have been loaded. The proposed change to `Condition*` keywords will result in silently skipping the unit if ipmi drivers aren't loaded (yet). Am I missing something here?

Also, since you've created ipmitool#46, please add 'Resolves ipmitool/ipmitool#46' to your commit message.

charles-rose:
> As far as I understand, the current implementation only runs this `exchange-bmc-os-info` unit after the ipmi drivers are known to have been loaded. The proposed change to `Condition*` keywords will result in silently skipping the unit if ipmi drivers aren't loaded (yet). Am I missing something here?
> 
Sorry for the late response.
I agree that the unit will fail without error. But the kernel modules autoload fairly early in the boot process. The motive was a to eliminate a dependency on a service (and package) that is mostly not necessary for systems with IPMI KCS.

If you believe keeping the Condition* would be better, I am fine with that as well.

> Also, since you've created #46, please add `Resolves ipmitool/ipmitool#46` to your commit message.

charles-rose:
Will suggest a better patch that lets systemd report failure when IPMI is not available.

PR #48, opened 2018-11-05
Correct segment violation in memmove when DCMI request returns C6.

pjdhpe:
On occasion a dcmi power reading will return error C6, and the decrypted payload is shorter than the expected length. The lanplus_decrypt_aes_cbc_128() adjusts the payload_size downward by one byte. In ipmi_lan_poll_single() a calculation to determine if the payload size has increased  results in the extra_data_length becoming -1, with a subsequent segv when calling a memmove to shift response data. The fix is to check for a positive value in the extra_data_length. The IPMI_PAYLOAD_TYPE_SOL further down the function has a similar fix, checking for a positive extra_data_length before calling the memmove..


AlexanderAmelkin:
You can `git push -f` to the same branch to amend the commit message, no need to close/re-submit.

Issue #13, opened 2018-05-31
Get rid of compilation warnings and set warnings as errors


AlexanderAmelkin:
There are myriads of warnings produced during compilation.
In order to make this tool behave more predictably and for overall cleanliness of the code it is required to thoughtfully and thoroughly get rid of all the warnings and prohibit introduction of new ones in future by treating warnings as errors in the compiler (`-Werror`).

PR #49, opened 2018-11-05
Cleanup compiler warnings

vmauery:
I really don't like seeing so many compiler warnings when ipmitool builds. This is the first round of removing a few of them.

### DIFF
--- a/lib/helper.c
+++ b/lib/helper.c
@@ -823,6 +823,7 @@ ipmi_start_daemon(struct ipmi_intf *intf)
 {
 	pid_t pid;
 	int fd;
+	int ret;
 #ifdef SIGHUP
 	sigset_t sighup;
 #endif
@@ -866,7 +867,11 @@ ipmi_start_daemon(struct ipmi_intf *intf)
 		exit(0);
 #endif
 
-	chdir("/");
+	ret = chdir("/");
+	if (ret) {
+		lprintf(LOG_ERR, "chdir failed: %s (%d)", strerror(errno), errno);
+		exit(1);
+	}
 	umask(0);
 
 	for (fd=0; fd<64; fd++) {
@@ -875,9 +880,20 @@ ipmi_start_daemon(struct ipmi_intf *intf)
 	}
 
 	fd = open("/dev/null", O_RDWR);
-	assert(0 == fd);
-	dup(fd);
-	dup(fd);
+	if (fd != STDIN_FILENO) {
+		lprintf(LOG_ERR, "failed to reset stdin: %s (%d)", strerror(errno), errno);
+		exit(1);
+	}
+	ret = dup(fd);
+	if (ret != STDOUT_FILENO) {
+		lprintf(LOG_ERR, "failed to reset stdout: %s (%d)", strerror(errno), errno);
+		exit(1);
+	}
+	ret = dup(fd);
+	if (ret != STDOUT_FILENO) {
+		lprintf(LOG_ERR, "failed to reset stderr: %s (%d)", strerror(errno), errno);
+		exit(1);
+	}
 }
 
 /* eval_ccode - evaluate return value of _ipmi_* functions and print error error
--- a/lib/ipmi_sdr.c
+++ b/lib/ipmi_sdr.c
@@ -2119,15 +2119,13 @@ ipmi_sdr_print_sensor_eventonly(struct ipmi_intf *intf,
 
 /* ipmi_sdr_print_sensor_mc_locator  -  print SDR MC locator record
  *
- * @intf:	ipmi interface
  * @mc:		mc locator sdr record
  *
  * returns 0 on success
  * returns -1 on error
  */
 int
-ipmi_sdr_print_sensor_mc_locator(struct ipmi_intf *intf,
-				 struct sdr_record_mc_locator *mc)
+ipmi_sdr_print_sensor_mc_locator(struct sdr_record_mc_locator *mc)
 {
 	char desc[17];
 
@@ -2215,15 +2213,13 @@ ipmi_sdr_print_sensor_mc_locator(struct ipmi_intf *intf,
 
 /* ipmi_sdr_print_sensor_generic_locator  -  print generic device locator record
  *
- * @intf:	ipmi interface
  * @gen:	generic device locator sdr record
  *
  * returns 0 on success
  * returns -1 on error
  */
 int
-ipmi_sdr_print_sensor_generic_locator(struct ipmi_intf *intf,
-				      struct sdr_record_generic_locator *dev)
+ipmi_sdr_print_sensor_generic_locator(struct sdr_record_generic_locator *dev)
 {
 	char desc[17];
 
@@ -2272,15 +2268,13 @@ ipmi_sdr_print_sensor_generic_locator(struct ipmi_intf *intf,
 
 /* ipmi_sdr_print_sensor_fru_locator  -  print FRU locator record
  *
- * @intf:	ipmi interface
  * @fru:	fru locator sdr record
  *
  * returns 0 on success
  * returns -1 on error
  */
 int
-ipmi_sdr_print_sensor_fru_locator(struct ipmi_intf *intf,
-				  struct sdr_record_fru_locator *fru)
+ipmi_sdr_print_sensor_fru_locator(struct sdr_record_fru_locator *fru)
 {
 	char desc[17];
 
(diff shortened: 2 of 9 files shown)


include/ipmitool/helper.h
```diff
@@ -38,6 +38,12 @@
 #include <stdio.h>
 #include <string.h>
 
+#ifdef __GNUC__
+#  define __UNUSED__(x) __attribute__((__unused__)) x
```
AlexanderAmelkin: Doesn't look like this new macro is used anywhere within this pull request.
Did you forget to replace occurrences of `__attribute__((__unused__))` with it?

include/ipmitool/helper.h
```diff
@@ -183,11 +189,23 @@ uint8_t *array_letoh(uint8_t *buffer, size_t length);
 #endif
 
 #ifndef __minlen
-# define __minlen(a, b) ({ int x=strlen(a); int y=strlen(b); (x < y) ? x : y;})
+static inline size_t min_strlen(const char *a, const char *b)
```
AlexanderAmelkin: Yup. Already have this done in my local branch.

include/ipmitool/ipmi_intf.h
```diff
@@ -58,7 +58,7 @@ enum LANPLUS_SESSION_STATE {
 };
 
 
-#define IPMI_AUTHCODE_BUFFER_SIZE 20
+#define IPMI_AUTHCODE_BUFFER_SIZE 20u
```
AlexanderAmelkin: What exactly are you addressing with this change? Doesn't look to affect the number of warnings.

src/plugins/ipmi_intf.c
```diff
@@ -219,7 +219,7 @@ ipmi_intf_session_set_username(struct ipmi_intf * intf, char * username)
 	if (!username)
 		return;
 
-	memcpy(intf->ssn_params.username, username, __min(strlen(username), 16));
+	memcpy(intf->ssn_params.username, username, __min(strlen(username), 16u));
```
AlexanderAmelkin: Just `16` without `u` is pretty well a `size_t` value. What's the need for `u` ? There were no warnings in ipmi_intf.c even without this change.

lib/helper.c
```diff
@@ -866,7 +867,8 @@ ipmi_start_daemon(struct ipmi_intf *intf)
 		exit(0);
 #endif
 
-	chdir("/");
+	ret = chdir("/");
+	assert(ret == 0);
```
AlexanderAmelkin: https://github.com/ipmitool/ipmitool/wiki/Coding-Standards
> Assertions
>
> `assert()` - don't; really, why would you?

Indeed, failing on an assertion is highly non-informative. Please properly check for the error and print the diagnostic message based on `errno`.

lib/helper.c
```diff
@@ -875,9 +877,11 @@ ipmi_start_daemon(struct ipmi_intf *intf)
...
 
 	fd = open("/dev/null", O_RDWR);
-	assert(0 == fd);
-	dup(fd);
-	dup(fd);
+	assert(STDIN_FILENO == fd);
+	ret = dup(fd);
+	assert(STDOUT_FILENO == ret);
+	ret = dup(fd);
+	assert(STDERR_FILENO == ret);
```
AlexanderAmelkin: Same here. Don't use assertions. Check return values of `dup()` and print out a errno-based message on errors.

lib/ipmi_sdr.c
```diff
@@ -204,7 +204,9 @@ sdr_convert_sensor_reading(struct sdr_record_full_sensor *sensor, uint8_t val)
 	case 1:
 		if (val & 0x80)
 			val++;
-		/* Deliberately fall through to case 2. */
+		result = (double) (((m * (int8_t) val) +
+				    (b * pow(10, k1))) * pow(10, k2));
+		break;
```
AlexanderAmelkin: Why? The comment is totally valid from gcc's point of view and correctly silences the implicit fall through warning.
See https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough

You here add unnecessary code duplication. There were no warnings in regard to this in ipmi_sdr.c. At least with gcc 5.4.0 on my Linux Mint 18.3 (Ubuntu Xenial). Wasn't it so on your system? What compiler?

lib/ipmi_sdr.c
```diff
@@ -284,7 +286,8 @@ sdr_convert_sensor_hysterisis(struct sdr_record_full_sensor *sensor, uint8_t val
 	case 1:
 		if (val & 0x80)
 			val++;
-		/* Deliberately fall through to case 2. */
+		result = (double) (((m * (int8_t) val) ) * pow(10, k2));
+		break;
```
AlexanderAmelkin: same here

lib/ipmi_sdr.c
```diff
@@ -359,7 +362,8 @@ sdr_convert_sensor_tolerance(struct sdr_record_full_sensor *sensor, uint8_t val)
 	case 1:
 		if (val & 0x80)
 			val++;
-		/* Deliberately fall through to case 2. */
+		result = (double) (((m * ((double)((int8_t) val)/2))) * pow(10, k2));
+		break;
```
AlexanderAmelkin: and here

include/ipmitool/helper.h
vmauery: Oops. I was originally just going to mark the unused parameters as unused, but then I decided to just remove them. I will remove this.

include/ipmitool/helper.h
vmauery: Ok, I will remove this.

include/ipmitool/ipmi_intf.h
vmauery: By default, any literals are type int. This makes the literal an unsigned int, which then makes any comparisons with size_t safe (or at least not complain about signed/unsigned warnings)

lib/helper.c
vmauery: sure.

lib/helper.c
vmauery: ok

lib/ipmi_sdr.c
vmauery: GCC removes the duplicate code and it gets rid of a warning. I am using gcc 7.3.0-27ubuntu1~18.04 on Ubuntu 18.04.

src/plugins/ipmi_intf.c
vmauery: Yes, it is not needed. It is just pedantically correct :)

include/ipmitool/ipmi_intf.h
AlexanderAmelkin: First of all this doesn't seem to actually remove any warnings. Secondly, I can't get my `gcc` to produce a warning when comparing an unsigned variable to a positive literal, can you? Lastly, if we go this way, we need to do it uniformly everywhere for all inherently unsigned literals. 
All in all, so far this looks like a superfluous change.

src/plugins/ipmi_intf.c
AlexanderAmelkin: Let's not touch code without need. That hinders future blaming.

lib/ipmi_sdr.c
AlexanderAmelkin: The concern about duplicate code is not that it occupies space. It is that it is a source of bugs. It is not uncommon when someone changes the to-be-identical code in one place and then forgets to change it in the rest of the copies.

The existing comment indeed does not work. However, even gcc 8.1.0 is happy with a `/* fall through */` comment. So please just replace it with `/* fall through */`.

vmauery: I am all for code depuplication. I also hate warnings when code compiles. I am of a mind of either fix the warnings or what is the point of having the compiler warn you? Really what I want is a hammer to tell the compiler that I really know what I am doing in this case to make it shut up. It looks like /* fall through */ does just that. I was not aware of this trick.

include/ipmitool/ipmi.h
```diff
@@ -162,7 +162,7 @@ struct ipmi_rs {
 	 * Looks like this is the length of the entire packet, including the RMCP
 	 * stuff, then modified to be the length of the extra IPMI message data
 	 */
-	int data_len;
+	size_t data_len;
```
AlexanderAmelkin: This change introduces like 18 new warnings. Please address them.

include/ipmitool/ipmi_intf.h
AlexanderAmelkin: Please remove this change unless you can show that it indeed helps in any way. As I said, I don't see any related warnings and can't even reproduce one 'in-vitro'.

include/ipmitool/ipmi.h
AlexanderAmelkin: I know. That's why there is #13 that is still not done yet...

PR #58, opened 2018-11-14
Cleanup all unused-parameter warnings

pstrinkle:
Add a .clang-format style definition to allow automated formatting with
clang-format.

Can run against files with changes to .clang-format via:

```
CLANG_FORMAT="clang-format-6.0"
find . \( -regextype sed -regex ".*\.[hc]\(pp\)\?"  \) \
    -not -name "*mako*" -not -type d -print0 |\
    xargs -0 "${CLANG_FORMAT}" -i
```


pstrinkle:
I didn't try building this first due to the nature of the change and a trust in clang-format-6.0
I'll give it a whirl and see what's going on.  I could also reduce the scope of the .clang-format s.t. it only fixes spaces, indentation and 80-char width, which may be sufficient.

I haven't used github PRs in a long time, so I'm less certain how to update them, versus bumping patchsets in gerrit (will take a look at some guides).

AlexanderAmelkin:
Haven't checked it yet, but I suspect it will not match https://github.com/ipmitool/ipmitool/wiki/Coding-Standards. Or will it?

pstrinkle:
Alexander, I force squashed the initial bad patch and the revert, and now it's only this one patch --, i'll set up a separate PR for other pieces.

lib/ipmi_delloem.c
```diff
@@ -1338,7 +1338,7 @@ ipmi_lcd_set_single_line_text(struct ipmi_intf * intf, char * text)
  *                  0 if successful
  */
 static int
-ipmi_lcd_set_text(struct ipmi_intf * intf, char * text, int line_number)
+ipmi_lcd_set_text(struct ipmi_intf * intf, char * text, int line_number __attribute__((unused)))
```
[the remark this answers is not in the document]
pstrinkle: Right, this builds on other systems -- Will fix.  Also, will examine each individual method to determine whether the parameters are for consistency of calling, or due to a function pointer signature.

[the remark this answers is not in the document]
pstrinkle: Ok, so that doesn't exist in helper.h, so you're asking me to add it, and then use it.  I'm going to stage a couple patches in this PR:
1) add macro
2) fix all the places it wasn't already used (plenty of __attribute__((unused)) in the codebase)
3) then fix my patchset to use it.

pstrinkle:
The state should now be ready, all three action items have been resolved.

lib/ipmi_sol.c
```diff
@@ -648,10 +648,12 @@ ipmi_print_sol_info(struct ipmi_intf * intf, uint8_t channel)
  *
  * function will return -1 if value is not valid, or
  * will return 0 if valid.
+ *
+ * base parameter currently unused, left in for extension.
  */
 int ipmi_sol_set_param_isvalid_uint8_t( const char *strval,
 					const char *name,
-					int base,
+					int __UNUSED__(base),
```
AlexanderAmelkin: Why did you decide to leave it? It's apparently a left-over from incomplete commit 94c0bd0a that replaced `strtol()` with `str2uint()` but forgot to remove the unused argument.

src/plugins/lan/auth.c
```diff
@@ -126,7 +126,16 @@ uint8_t * ipmi_auth_md5(struct ipmi_session * s, uint8_t * data, int data_len)
  * Use OpenSSL implementation of MD2 algorithm if found.
  * This function is analogous to ipmi_auth_md5
  */
-uint8_t * ipmi_auth_md2(struct ipmi_session * s, uint8_t * data, int data_len)
+uint8_t * ipmi_auth_md2(
+#ifdef HAVE_CRYPTO_MD2
```
AlexanderAmelkin: To reduce duplication, please define a macro like:
```
#ifdef HAVE_CRYPTO_MD2
  #define __MD2_ONLY__(x) x
#else 
  #deine __MD2_ONLY__(x) __UNUSED__(x)
#endif 
```
and use it with the arguments of the function. That way you will have each argument declared only once.

src/plugins/lan/lan.c
```diff
@@ -338,7 +338,7 @@ ipmi_lan_recv_packet(struct ipmi_intf * intf)
  * asf.data[f:a]= 0x000000000000
  */
 static int
-ipmi_handle_pong(struct ipmi_intf * intf, struct ipmi_rs * rsp)
+ipmi_handle_pong(struct ipmi_rs * rsp)
```
AlexanderAmelkin: Asterisks should stick to the variable name as per [Coding Standards](https://github.com/ipmitool/ipmitool/wiki/Coding-Standards#pointers). I know it wasn't you who put it this way, but once you touch the line, please fix it. Applies everywhere.

pstrinkle:
There was one static array in the header used by two source files, so I could mark it extern in the header and store it in one of the files -- however, IIRC, the C compiler isn't as fond of doing that without a fixed array length -- but it's been a while.

PR #60, opened 2018-11-24
Cleanup ipmi_sel module by splitting out OEM pieces

pstrinkle:
Split out the OEM pieces from the ipmi_sel module.


pstrinkle:
Two methods moved are now ranked complex.  Alexander, thoughts on this?  I was hoping to avoid re-working the Dell SEL code in this PR.  Even though, it definitely needs some TLC.

pstrinkle:
I'm going to split up the functions, I read through the dell one, and it's pretty straightforward in its goal, and looks easy to split via helper methods to handle case-specific items.

PR #63, opened 2018-11-26
Address two memory leaks

pstrinkle:
Addresses two memory leaks.  Before fixing each leak, I fixed the style within that specific method.  If you approve of the style changes in a different PR, I can start applying that more liberally.


lib/ipmi_sol.c
```diff
@@ -1576,110 +1570,92 @@ ipmi_sol_red_pill(struct ipmi_intf * intf, int instance)
...
 
-		if (retval)
-		{
-			if (retval == -1)
-			{
+		if (retval) {
+			if (retval == -1) {
 				/* ERROR */
 				perror("select");
+				free(buffer);
```
pstrinkle: I wasn't sure if they needed to explicitly leave raw mode before exiting -- if so, this can be added here, or this can jump to a cleanup label.

pstrinkle:
Alexander, fixed the static bit on these two files.  Other cleanup while I had it open.

pstrinkle:
There are some memory leak bugfixes here, and then some reworked pieces as you desired.  However, it does have a lot of the automated whitespace cleanup.  I can spend a bit and rebase all the patches and exclude the whitespace fix patches.

PR #65, opened 2018-11-26
bugfix: ipmi_fru: fixup array bounds checking

pstrinkle:
Fixup the following array bounds checking bugs:
[lib/ipmi_fru.c:1003]: (style) Array index 'i' is
used before limits check.
[lib/ipmi_fru.c:1127]: (style) Array index 'i' is
used before limits check.
[lib/ipmi_fru.c:1262]: (style) Array index 'i' is
used before limits check.


pstrinkle:
Rebased the changes and ended up with a separate PR.

PR #66, opened 2018-11-26
bugfix: lib/ipmi_ekanalyzer: fixup resource leak

pstrinkle:
fixup a resource leak in lib/ipmi_ekanalyzer.c, and fixup the style surrounding the bugfix.


pstrinkle:
Updated free calls, but, in this file, there is a lot of style work required to bring it up to snuff.  I'd like to save that for another PR set down the line.

PR #67, opened 2018-11-26
cleanup: lib/ipmi_sunoem: drop ineffective variable assignment

pstrinkle:
Fix the following:
[lib/ipmi_sunoem.c:203]: (warning) Assignment of
function parameter has no effect outside the function.


pstrinkle:
Fixed up the static and return value bits.

PR #69, opened 2018-11-26
bugfix: lib/ipmi_hpmfwupg: fixup resource leak

pstrinkle:
Cleanup the function and then fixup a resource leak.


pstrinkle:
It looks like the resource leak this PR was fixing as been fixed.  I'll circle back to see if this is a useful PR anymore.

PR #70, opened 2018-11-26
style: lib/ipmi_sdr: remove unused assignments

pstrinkle:
Remove assignments with no effect:
[lib/ipmi_sdr.c:3190]:
[lib/ipmi_sdr.c:3242]:
(warning) Assignment of function parameter has no
effect outside the function. Did you forget
dereferencing it?

Also fixup the assignment to e within the for loop, as the iterative
step is immediately overwriting this assignment with f.


pstrinkle:
Like the other PRs, I'm updating them today to split out the changes more cleanly, and also to remove places where I deleted code that is specified as "should be there."

pstrinkle:
Updated static returns -- there are some other bits where the code could be improved by using helper methods -- will evaluate on demand.

PR #73, opened 2018-11-27
cleanup: lib/ipmi_sensor.c

pstrinkle:
Deliberately cast return values to void when ignoring.
[lib/ipmi_sensor.c:712, 724, 748, 760]: (style)
Variable 'ret' is reassigned a value before the old
one has been used.
[lib/ipmi_sensor.c:791]: (style) Variable 'rsp' is
reassigned a value before the old one has been used.

Cleanup the style within the impacted methods and simplify logic with
helper functions.


pstrinkle:
Fixed static method definitions impacted by this PR.

PR #74, opened 2018-11-27
[WIP] cleanup: lipmi

pstrinkle:
[src/plugins/lipmi/lipmi.c:78]: (style) Variable
'curr_seq' is assigned a value that is never used.


pstrinkle:
Updated static returns, per style guide.

PR #76, opened 2018-11-27
cleanup: src/plugins/lanplus/lanplus.c

pstrinkle:


src/plugins/lanplus/lanplus.c
```diff
@@ -2163,25 +2161,22 @@ ipmi_lanplus_send_payload(
...
 				{
-					lprintf(LOG_DEBUG+1, "BUILDING A v1.5 COMMAND");
+					lprintf(LOG_DEBUG + 1, "BUILDING A v1.5 COMMAND");
 					entry = ipmi_lanplus_build_v15_ipmi_cmd(intf, ipmi_request);
-				}
-				else
-				{
-					int isRetry = ( try > 0 ? 1 : 0 );
+				} else {
+					int isRetry = (try > 0) ? 1 : 0;
```
AlexanderAmelkin: This was a programming pleonasm. Once you're touching it, please keep just `(try > 0)`.
It woul also be nice to change isRetry to `bool` with `#include <stdbool.h>`.

src/plugins/lanplus/lanplus.c
```diff
@@ -2366,7 +2339,6 @@ ipmi_lanplus_send_payload(
 	case IPMI_PAYLOAD_TYPE_RAKP_3:
 	case IPMI_PAYLOAD_TYPE_SOL:
 		free(msg_data);
-		msg_data = NULL;
```
AlexanderAmelkin: Please don't remove such assignments unless immediately before the end of their visibility scope.
According to C standards, accessing an invalid pointer (not the data it points to, but the pointer itself) yields an undefined behavior. Besides that, it is impossible to tell a valid pointer from an invalid one unless it's NULL. Hence, there is [this](https://github.com/ipmitool/ipmitool/wiki/Coding-Standards#memory-allocation) rule.
This line is not __immediately__ before the end of scope, so please leave the assignment.

FYR, https://stackoverflow.com/questions/17024866/when-is-it-valid-to-access-a-pointer-to-a-dead-object

src/plugins/lanplus/lanplus.c
```diff
@@ -603,11 +603,11 @@ ipmiv2_lan_ping(struct ipmi_intf * intf)
  * Returns: the ipmi_rs packet describing the/a response we expect.
  */
 static struct ipmi_rs *
-ipmi_lan_poll_single(struct ipmi_intf * intf)
+ipmi_lan_poll_single(struct ipmi_intf *intf)
```
AlexanderAmelkin: I really think that formatting changes must not be confused in the same commit with code changes like removing the unused variables. It's quite hard to review.

src/plugins/lanplus/lanplus.c
pstrinkle: I'm surprised by the explicit ptr = NULL after free when you can trace the code to see it's never actually checked against NULL before it's then re-allocated.

src/plugins/lanplus/lanplus.c
pstrinkle: Yeah, originally I was keeping them quite separate, however you've instructed me to fix every line I touch and then the surrounding lines as well, it just became a bit of a back and forth to have a style fix commit, then a real fix commit, back and forth for a given code block.

I did push a PR that will deal with style that requires your feedback, if that's used, I can fix up all the files and then rebase these patches removing the extra noise.

src/plugins/lanplus/lanplus.c
pstrinkle: Ack

pstrinkle:
Alexander, PTAL - I rebased these patches onto a style commit to remove the style changes from each patch.

src/plugins/lanplus/lanplus.c
```diff
@@ -1364,56 +1341,61 @@ void getIpmiPayloadWireRep(
...
+
+		msg[len++] = 0x34; /* Send Message rqst */
+		if (bridgedRequest == 2)
+			msg[len++] =
+				(0x40
+				 | intf->transit_channel); /* Track request*/
 		else
-			msg[len++] = (0x40|intf->target_channel);    /* Track request*/
-			
+			msg[len++] =
```
pstrinkle: This can probably be cleaned by moving the comment from the end of the line to immediately above and adding curly braces.

src/plugins/lanplus/lanplus.c
```diff
@@ -1449,34 +1431,32 @@ void getIpmiPayloadWireRep(
...
 /*
  * getSolPayloadWireRep
  *
  * param msg [out] will contain our wire representation
  * param payload [in] holds the v2 payload with our SOL data
  */
-void getSolPayloadWireRep(
-						  uint8_t          * msg,     /* output */
-						  struct ipmi_v2_payload * payload) /* input */
+void getSolPayloadWireRep(uint8_t *msg,			   /* output */
```
pstrinkle: I don't think clang-format is happy with parameter comments, because it didn't seem to clean them all consistently.

src/plugins/lanplus/lanplus.c
```diff
@@ -897,58 +904,57 @@ ipmi_lan_poll_recv(struct ipmi_intf * intf)
...
+#if WORDS_BIGENDIAN
+	rsp->payload.open_session_response.bmc_id =
+		BSWAP_32(rsp->payload.open_session_response.bmc_id);
+#endif
+
+	/* And of course, our negotiated algorithms */
+	rsp->payload.open_session_response.auth_alg = rsp->data[offset + 16];
+	rsp->payload.open_session_response.integrity_alg =
+		rsp->data[offset + 24];
+	rsp->payload.open_session_response.crypt_alg = rsp->data[offset + 32];
```
AlexanderAmelkin: Need to do something with these magic offsets. As well as with the `8` and `4` above.
Besides, the old formatting was much better readable. I prefer violating the line length over making the code less readable.

src/plugins/lanplus/lanplus.c
```diff
@@ -1056,39 +1057,35 @@ read_rakp2_message(
...
 
-	 /* RAKP response code */
-	 rsp->payload.rakp4_message.rakp_return_code = rsp->data[offset + 1];
+	/* RAKP response code */
+	rsp->payload.rakp4_message.rakp_return_code = rsp->data[offset + 1];
 
 	/* Console session ID */
-	rsp->payload.rakp4_message.console_id = ipmi32toh(&rsp->data[offset + 4]);
+	rsp->payload.rakp4_message.console_id =
+		ipmi32toh(&rsp->data[offset + 4]);
```
AlexanderAmelkin: Magic offsets...

src/plugins/lanplus/lanplus.c
```diff
@@ -1056,39 +1057,35 @@ read_rakp2_message(
...
+	case IPMI_AUTH_RAKP_NONE:
+		/* Nothing to do here */
+		break;
 
-	 case IPMI_AUTH_RAKP_HMAC_SHA1:
+	case IPMI_AUTH_RAKP_HMAC_SHA1:
 		/* We need to copy 12 bytes */
 		for (i = 0; i < IPMI_SHA1_AUTHCODE_SIZE; ++i) {
 			rsp->payload.rakp4_message.integrity_check_value[i] =
 				rsp->data[offset + 8 + i];
```
AlexanderAmelkin: More magic offsets

src/plugins/lanplus/lanplus.c
```diff
@@ -1131,10 +1128,7 @@ read_rakp4_message(
  *               this function is called.  The offset will be adjusted to
  *               point to the end of the session when this function exits.
  */
-void
-read_session_data(
-				  struct ipmi_rs * rsp,
-				  int * offset)
+void read_session_data(struct ipmi_rs *rsp, int *offset)
```
AlexanderAmelkin: https://github.com/ipmitool/ipmitool/wiki/Coding-Standards#function-definition

src/plugins/lanplus/lanplus.c
```diff
@@ -1165,14 +1158,11 @@ read_session_data(
  *              the end of the session when this function exits.
  *  param s      holds our session state
  */
-void
-read_session_data_v2x(
-					  struct ipmi_rs      * rsp,
-					  int                 * offset)
+void read_session_data_v2x(struct ipmi_rs *rsp, int *offset)
```
AlexanderAmelkin: https://github.com/ipmitool/ipmitool/wiki/Coding-Standards#function-definition

src/plugins/lanplus/lanplus.c
```diff
@@ -1165,14 +1158,11 @@ read_session_data(
...
 {
 	rsp->session.authtype = rsp->data[(*offset)++];
 
-	rsp->session.bEncrypted     = (rsp->data[*offset] & 0x80 ? 1 : 0);
+	rsp->session.bEncrypted = (rsp->data[*offset] & 0x80 ? 1 : 0);
 	rsp->session.bAuthenticated = (rsp->data[*offset] & 0x40 ? 1 : 0);
```
AlexanderAmelkin: Magic masks this time... Definitely needs a separate commit to get rid of 'em all.

src/plugins/lanplus/lanplus.c
```diff
@@ -1209,17 +1198,16 @@ read_session_data_v2x(
  *              end of the session when this function exits.
  * param s      holds our session state
  */
-void read_session_data_v15(
-							struct ipmi_rs * rsp,
-							int * offset)
+void read_session_data_v15(struct ipmi_rs *rsp, int *offset)
```
AlexanderAmelkin: https://github.com/ipmitool/ipmitool/wiki/Coding-Standards#function-definition

src/plugins/lanplus/lanplus.c
```diff
@@ -1209,17 +1198,16 @@ read_session_data_v2x(
...
 {
 	/* All v15 messages are IPMI messages */
 	rsp->session.payloadtype = IPMI_PAYLOAD_TYPE_IPMI;
 
 	rsp->session.authtype = rsp->data[(*offset)++];
 
-	/* All v15 messages that we will receive are unencrypted/unauthenticated */
-	rsp->session.bEncrypted     = 0;
+	/* All v15 messages that we will receive are unencrypted/unauthenticated
+	 */
```
AlexanderAmelkin: Really?

src/plugins/lanplus/lanplus.c
AlexanderAmelkin: I guess that must be done then.

src/plugins/lanplus/lanplus.c
AlexanderAmelkin: Maybe clang-format is just the wrong tool for this project? Don't you think so?

src/plugins/lanplus/lanplus.c
```diff
@@ -1868,7 +1869,7 @@ ipmi_lanplus_build_v2x_ipmi_cmd(struct ipmi_intf *intf, struct ipmi_rq *req,
 	 */
 	static uint8_t curr_seq = 0;
 
-	if (isRetry == 0)
+	if (isRetry == false)
```
AlexanderAmelkin: `if (!isRetry)`

src/plugins/lanplus/lanplus.c
```diff
@@ -1868,7 +1869,7 @@ ipmi_lanplus_build_v2x_ipmi_cmd(struct ipmi_intf *intf, struct ipmi_rq *req,
...
 		curr_seq += 1;
 
 	if (curr_seq >= 64)
```
AlexanderAmelkin: What is `64`, by the way? Do you know? Looks like some maximum possible sequence number (probably from IPMI spec).

src/plugins/lanplus/lanplus.c
```diff
@@ -2048,7 +2048,7 @@ ipmi_lanplus_build_v15_ipmi_cmd(struct ipmi_intf *intf, struct ipmi_rq *req)
 /*
  * is_sol_packet
  */
-static int is_sol_packet(struct ipmi_rs *rsp)
+static bool is_sol_packet(struct ipmi_rs *rsp)
```
AlexanderAmelkin: Types on separate lines. https://github.com/ipmitool/ipmitool/wiki/Coding-Standards#function-definition

src/plugins/lanplus/lanplus.c
AlexanderAmelkin: It's not because it's checked before reallocation. It's to make sure it is NULL because otherwise you can't tell a valid pointer from an invalid one and as explained on StackOverflow, even accessing a pointer itself may lead to undefined behavior if it's invalid.

PR #77, opened 2018-11-27
[WIP] cleanup: vita: cleanup some bit checks

pstrinkle:
[lib/ipmi_vita.c:464, 466, 468]: (style) Clarify
calculation precedence for '&' and '?'.

### DIFF
--- a/include/ipmitool/helper.h
+++ b/include/ipmitool/helper.h
@@ -120,6 +120,17 @@ uint16_t ipmi_get_oem_id(struct ipmi_intf *intf);
 
 #define IS_SET(v, b) ((bool)((v) & (1 << (b))))
 
+/**
+ * Free the memory and clear the pointer.
+ * @param[in] ptr - a pointer to your pointer to free.
+ */
+static inline void free_n(void **ptr) {
+	if (ptr && *ptr) {
+		free(*ptr);
+		*ptr = NULL;
+	}
+}
+
 /* le16toh(), hto16le(), et. al. don't exist for Windows or Apple */
 /* For portability, let's simply define our own versions here */
 
--- a/include/ipmitool/ipmi_fru.h
+++ b/include/ipmitool/ipmi_fru.h
@@ -42,6 +42,8 @@
 # include <config.h>
 #endif
 
+#define FRU_END_OF_FIELDS 0xc1
+
 #define GET_FRU_INFO		0x10
 #define GET_FRU_DATA		0x11
 #define SET_FRU_DATA		0x12
@@ -587,6 +589,11 @@ struct fru_picmgext_amc_link_desc_record {
 #ifdef HAVE_PRAGMA_PACK
 #pragma pack(0)
 #endif
+
+/* IPMI Return codes for Get FRU Inventory Area and Write FRU Inventory Area */
+/* PROTECTED_OFFSET Only expected on write command failures. */
+#define IPMI_CC_FRU_WRITE_PROTECTED_OFFSET 0x80
+#define IPMI_CC_FRU_DEVICE_BUSY 0x81
 
 /* FRU Board manufacturing date */
 #define FRU_BOARD_DATE_UNSPEC 0 /* IPMI FRU Information Storage Definition
--- a/lib/ipmi_fru.c
+++ b/lib/ipmi_fru.c
@@ -33,6 +33,7 @@
 #include <ipmitool/ipmi.h>
 #include <ipmitool/log.h>
 #include <ipmitool/helper.h>
+#include <ipmitool/ipmi_cc.h>
 #include <ipmitool/ipmi_intf.h>
 #include <ipmitool/ipmi_fru.h>
 #include <ipmitool/ipmi_mc.h>
@@ -40,6 +41,7 @@
 #include <ipmitool/ipmi_strings.h>  /* IANA id strings */
 #include <ipmitool/ipmi_time.h>
 
+#include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
@@ -98,6 +100,12 @@ static const char * chassis_type_desc[] = {
 	"Blade",
 	"Blade Enclosure"
 };
+
+static inline bool fru_cc_rq2big(int code) {
+	return (code == IPMI_CC_REQ_DATA_INV_LENGTH
+		|| code == IPMI_CC_REQ_DATA_FIELD_EXCEED
+		|| code == IPMI_CC_CANT_RET_NUM_REQ_BYTES);
+}
 
 /* From lib/dimm_spd.c: */
 int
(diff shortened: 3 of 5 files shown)


pstrinkle:
I applied a general style fix commit, then the change, then the IS_SET change.

pstrinkle:
I don't see a button to indicate, "PTAL again, things should be fixed"

lib/ipmi_vita.c
```diff
@@ -357,7 +317,7 @@ ipmi_vita_get_vso_capabilities(struct ipmi_intf *intf)
 	tmp = (rsp->data[2] & 0x30) >> 4;
 
 	printf("    Frequency  %skHz\n",
-		tmp == 0 ? "100" : tmp == 1 ? "400" : "RESERVED");
+	       tmp == 0 ? "100" : tmp == 1 ? "400" : "RESERVED");
```
AlexanderAmelkin: Now this is what indeed needed clarification of precedence. Also, I'd format it like this for ease of comprehension:

```
printf("    Frequency %skHz\n",
       (tmp == 0)
       ? "100"
       : (tmp == 1)
         ? "400"
         : "RESERVED"
      );
```
Or better yet, get rid of the ternary operator and use a switch:

```
static const char const *rsvd = "RESERVED";
const char *freq = rsvd;

switch (tmp) {
case 0:
    freq = "100";
    break;
case 1:
    freq = "400";
    break;
}
```

You could then use `rsvd` in other places where `RESERVED` is printed and reduce data duplication a little.

lib/ipmi_vita.c
```diff
@@ -368,20 +328,19 @@ ipmi_vita_get_vso_capabilities(struct ipmi_intf *intf)
 	}
 
 	printf("VSO Standard      : %s\n",
-		(rsp->data[3] & 0x3) == 0 ? "VITA 46.11" : "RESERVED");
+	       (rsp->data[3] & 0x3) == 0 ? "VITA 46.11" : "RESERVED");
```
AlexanderAmelkin: Like here. I wish this `0x3` was a macro as well.

lib/ipmi_vita.c
```diff
@@ -368,20 +328,19 @@ ipmi_vita_get_vso_capabilities(struct ipmi_intf *intf)
...
 
 	printf("VSO Spec Revision : %d.%d\n", rsp->data[4] & 0xf,
-		rsp->data[4] >> 4);
+	       rsp->data[4] >> 4);
```
AlexanderAmelkin: And these `0xf` and `4`... That's a separate PR though, I guess.

lib/ipmi_vita.c
```diff
@@ -368,20 +328,19 @@ ipmi_vita_get_vso_capabilities(struct ipmi_intf *intf)
...
 	printf("FRU Device ID     : 0x%02x\n", rsp->data[6]);
 
 	return 0;
 }
 
-static int
-ipmi_vita_set_fru_activation(struct ipmi_intf *intf,
-	char **argv, unsigned char command)
+static int ipmi_vita_set_fru_activation(struct ipmi_intf *intf, char **argv,
+					unsigned char command)
```
AlexanderAmelkin: What a strange alignment of the second line!

lib/ipmi_vita.c
```diff
@@ -394,11 +353,11 @@ ipmi_vita_set_fru_activation(struct ipmi_intf *intf,
 	req.msg.data = msg_data;
 	req.msg.data_len = 3;
 
-	msg_data[0]	= GROUP_EXT_VITA;		/* VITA identifier */
-	if (is_fru_id(argv[0], &msg_data[1]) != 0) {	/* FRU ID */
+	msg_data[0] = GROUP_EXT_VITA;		     /* VITA identifier */
```
AlexanderAmelkin: Again, no need to pretty-format. Or we'll be fixing this forever.

lib/ipmi_vita.c
```diff
@@ -657,8 +614,8 @@ ipmi_vita_get_led_state(struct ipmi_intf *intf, char **argv)
 			val2str(rsp->ccode, completion_code_vals));
 		return -1;
 	} else if (rsp->data_len < 5
-		|| ((rsp->data[1] & 0x2) && rsp->data_len < 8)
-		|| ((rsp->data[1] & 0x4) && rsp->data_len < 9)) {
+		   || ((rsp->data[1] & 0x2) && rsp->data_len < 8)
+		   || ((rsp->data[1] & 0x4) && rsp->data_len < 9)) {
```
AlexanderAmelkin: *singing* Maaaagic... moooments... ;)

lib/ipmi_vita.c
```diff
@@ -950,7 +903,7 @@ ipmi_vita_main (struct ipmi_intf *intf, int argc, char **argv)
 	case VITA_CMD_POLICY_GET:
 		if (argc > 2) {
 			rc = ipmi_vita_get_fru_state_policy_bits(intf,
-				&argv[2]);
+								 &argv[2]);
```
AlexanderAmelkin: Could as well invert the condition and reduce indentation:
```
if (argc <= 2) {
    break;
}

rc = ipmi_vita...
break;
```

lib/ipmi_vita.c
```diff
@@ -39,136 +39,104 @@
...
+	 "               6:   WHITE\n"
+	 "               0xE: do not change\n"
+	 "               0xF: use default color"},
+	{VITA_CMD_UNKNOWN, "Unknown command"},
+	{0, NULL}};
 
 /* check if VITA 46.11 is supported */
-uint8_t
-vita_discover(struct ipmi_intf *intf)
+uint8_t vita_discover(struct ipmi_intf *intf)
```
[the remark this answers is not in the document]
pstrinkle: Yeah, didn't see this (https://github.com/ipmitool/ipmitool/wiki/Coding-Standards#function-definition), as just about none of the code follows the standard -- I'll adjust the clang-format.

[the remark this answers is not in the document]
pstrinkle: Also, I'm using the clang-format for https://github.com/openbmc/pam-ipmi/blob/master/.clang-format -- some defaults for values not explicitly set. -- do you have a clang-format file I can use?  I don't believe humans should be formatting code.  I spent years having to format my own code and it can such a time sync.

pstrinkle:
I'm just going to fix this file by hand, it's not super long.  As far as the magic numbers... another PR would be preferred so this doesn't carry on indefinitely.  there are quite a few files I haven't even hit -- the only ones I've hit were ones with weird code that was caught automatically or in the case of the open plugin, to apply a real bugfix, ..., also a few other bugfixes scattered.

lib/ipmi_vita.c
```diff
@@ -202,11 +171,9 @@ vita_discover(struct ipmi_intf *intf)
 		lprintf(LOG_INFO, "Invalid completion code received: %s",
 			val2str(rsp->ccode, completion_code_vals));
 	} else if (rsp->data_len < 5) {
-		lprintf(LOG_INFO, "Invalid response length %d",
-			rsp->data_len);
+		lprintf(LOG_INFO, "Invalid response length %d", rsp->data_len);
```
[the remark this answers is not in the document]
pstrinkle: clang-format tries to break as late as it can for parameters such that, in the top case it couldn't have the format string on the same line as the log level.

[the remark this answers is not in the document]
pstrinkle: In this case, it seems a bit wonky since it could have fit that line.

lib/ipmi_vita.c
pstrinkle: In this case, i just bumped the argv[2] onto the same line as the call.  It goes over 80 but it's close enough.

lib/ipmi_vita.c
pstrinkle: separate PR.

lib/ipmi_vita.c
pstrinkle: separate PR.

lib/ipmi_vita.c
pstrinkle: separate PR.

pstrinkle:
New comments all addressed.

pstrinkle:
I'll have to go through the other PRs and try to fix the static, return on different lines.

pstrinkle:
Per previous PR comments, this'll need to have its initial patch dropped.

PR #80, opened 2018-11-29
helper: add FREE helper macro

### LINKED ISSUE #79: Provide helper macro for freeing pointers
Either:

`#define FREE(PTR) do { free(PTR); (PTR) = NULL; } while(0);`

or
`void free_and_nullify(void **ptr) {
    if (ptr && *ptr) {
        free(*ptr);
        *ptr = NULL;
    }
}`

This will save developers from needing to set a pointer explicitly to null in cases where doing so is non-intuitive, e.g. at the end of a method.

pstrinkle:
Implement a free and nullify helper macro such
that developers need only call FREE(ptr) instead of

{
 free(ptr);
 ptr = NULL;
}

Partially resolves #79

### DIFF
--- a/include/ipmitool/helper.h
+++ b/include/ipmitool/helper.h
@@ -58,6 +58,12 @@
     #define __UNUSED__(x) x
 #endif
 
+#ifdef FREE
+#error "FREE already defined."
+#else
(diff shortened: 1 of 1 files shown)


AlexanderAmelkin:
Let's not merge it. See my comment in #79.

Issue #59, opened 2018-11-19
Need to get rid of bitfields in the code


AlexanderAmelkin:
Bit fields are completely non-portable as the order of bits within the unit 'implementation defined' as per the C11 standard, and is also 'ABI-specific' for at least gcc implementation. That is, we can't be sure how bit-fields are arranged even for a single compiler.

All bit-fields use cases must be changed to use mask and shift macros and/or inline functions.

This may also be partially related to #26.

PR #81, opened 2018-11-29
cleanup: fru

pstrinkle:
Various fixups for ipmi_fru.c

### DIFF
--- a/include/ipmitool/ipmi_fru.h
+++ b/include/ipmitool/ipmi_fru.h
@@ -588,6 +588,11 @@ struct fru_picmgext_amc_link_desc_record {
 #pragma pack(0)
 #endif
 
+/* IPMI Return codes for Get FRU Inventory Area and Write FRU Inventory Area */
+/* PROTECTED_OFFSET Only expected on write command failures. */
+#define IPMI_CC_FRU_WRITE_PROTECTED_OFFSET	0x80
+#define IPMI_CC_FRU_DEVICE_BUSY			0x81
+
 /* FRU Board manufacturing date */
 #define FRU_BOARD_DATE_UNSPEC 0 /* IPMI FRU Information Storage Definition
                                    v1.0 rev 1.3, Table 11-1 */
--- a/lib/ipmi_fru.c
+++ b/lib/ipmi_fru.c
@@ -1940,22 +1940,21 @@ static bool ipmi_fru_oemkontron_edit(int argc, char **argv, uint8_t *fru_data,
 }
 
 /* ipmi_fru_picmg_ext_edit  -  Query new values to replace original FRU content
-*
-* @data:   FRU data
-* @offset: start of the current multi record (start of header)
-* @len:    len of the current record (excluding header)
-* @h:      pointer to record header
-* @oh:     pointer to OEM /PICMG header
-*
-* returns: TRUE if data changed
-* returns: FALSE if data not changed
-*/
-static int ipmi_fru_picmg_ext_edit(uint8_t * fru_data,
-												int off,int len,
-												struct fru_multirec_header *h,
-												struct fru_multirec_oem_header *oh)
+ *
+ * @data:   FRU data
+ * @offset: start of the current multi record (start of header)
+ * @len:    len of the current record (excluding header)
+ * @h:      pointer to record header
+ * @oh:     pointer to OEM /PICMG header
+ *
+ * returns: TRUE if data changed
+ * returns: FALSE if data not changed
+ */
+static bool ipmi_fru_picmg_ext_edit(uint8_t *fru_data, int off, int len,
+				    struct fru_multirec_header *h,
+				    struct fru_multirec_oem_header *oh)
 {
-	int hasChanged = FALSE;
+	bool hasChanged = false;
 	int start = off;
 	int offset = start;
 	int length = len;
@@ -1975,12 +1974,13 @@ static int ipmi_fru_picmg_ext_edit(uint8_t * fru_data,
 				printf("      Maximum Internal Current(@12V): %.2f A (0x%02x)\n",
 								(float)max_current / 10.0f, max_current);
 
-				if( ipmi_fru_query_new_value(fru_data,index,2) ){
-					max_current = fru_data[index];
-					max_current |= fru_data[++index]<<8;
-					printf("      New Maximum Internal Current(@12V): %.2f A (0x%02x)\n",
-								(float)max_current / 10.0f, max_current);
-					hasChanged = TRUE;
+			if (ipmi_fru_query_new_value(fru_data, index, 2)) {
+				max_current = fru_data[index];
+				max_current |= fru_data[++index] << 8;
+				printf("      New Maximum Internal Current(@12V): %.2f A (0x%02x)\n",
+				       (float)max_current / 10.0f, max_current);
+				hasChanged = true;
+			}
 
 				}
 
@@ -2016,10 +2016,9 @@ static int ipmi_fru_picmg_ext_edit(uint8_t * fru_data,
 				if( ipmi_fru_query_new_value(fru_data, index, 1) ){
 					current = fru_data[index];
 
-					printf("      New Current draw(@12V): %.2f A (0x%02x)\n",
-								(float)current / 10.0f, current);
-					hasChanged = TRUE;
-				}
+				printf("      New Current draw(@12V): %.2f A (0x%02x)\n",
+				       (float)current / 10.0f, current);
+				hasChanged = true;
 			}
 			break;
 	}
--- a/lib/ipmi_fru.c
+++ b/lib/ipmi_fru.c
@@ -153,7 +153,9 @@ ipmi_fru_get_multirec_size_from_file(char *pFileName,
 				     uint32_t *pSize,
 				     uint32_t *pOffset);
 
-int ipmi_fru_get_adjust_size_from_buffer(uint8_t *pBufArea, uint32_t *pSize);
+static
+int
+ipmi_fru_get_adjust_size_from_buffer(uint8_t *pBufArea, uint32_t *pSize);
 
 static
 void
@@ -183,7 +185,9 @@ fru_area_print_multirec_bloc(struct ipmi_intf *intf,
 int read_fru_area(struct ipmi_intf *intf, struct fru_info *fru, uint8_t id,
 		  uint32_t offset, uint32_t length, uint8_t *frubuf);
 
-void free_fru_bloc(t_ipmi_fru_bloc *bloc);
+static
+void
+free_fru_bloc(t_ipmi_fru_bloc *bloc);
 
 /* get_fru_area_str  -  Parse FRU area string from raw data
 *
--- a/lib/ipmi_fru.c
+++ b/lib/ipmi_fru.c
@@ -3769,10 +3769,10 @@ ipmi_fru_get_multirec(struct ipmi_intf * intf, uint8_t id ,
 	return 0;
 }
 
-static int
-ipmi_fru_upg_ekeying(struct ipmi_intf * intf,
-			char * pFileName,
-			uint8_t fruId)
+#define ERR_EXIT do { rc = -1; goto exit; } while(0)
+
+static int ipmi_fru_upg_ekeying(struct ipmi_intf *intf, char *pFileName,
+				uint8_t fruId)
 {
 	struct fru_info fruInfo = {0};
 	uint8_t *buf = NULL;
@@ -3780,14 +3780,16 @@ ipmi_fru_upg_ekeying(struct ipmi_intf * intf,
 	uint32_t fruMultiRecSize = 0;
 	uint32_t offFileMultiRec = 0;
 	uint32_t fileMultiRecSize = 0;
+	int rc = 0;
+
 	if (!pFileName) {
 		lprintf(LOG_ERR, "File expected, but none given.");
-		return -1;
+		ERR_EXIT;
 	}
 	if (ipmi_fru_get_multirec_location_from_fru(intf, fruId, &fruInfo,
 							&offFruMultiRec, &fruMultiRecSize) != 0) {
 		lprintf(LOG_ERR, "Failed to get multirec location from FRU.");
-		return -1;
+		ERR_EXIT;
 	}
 	lprintf(LOG_DEBUG, "FRU Size        : %lu\n", fruMultiRecSize);
 	lprintf(LOG_DEBUG, "Multi Rec offset: %lu\n", offFruMultiRec);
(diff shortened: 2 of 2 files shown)


lib/ipmi_fru.c
```diff
@@ -850,8 +853,9 @@ int read_fru_area_section(struct ipmi_intf *intf, struct fru_info *fru,
 			/* if we get C7 or C8  or CA return code then we
 			 * requested too many bytes at once so try again with
 			 * smaller size */
-			if ((rsp->ccode == 0xc7 || rsp->ccode == 0xc8
-			     || rsp->ccode == 0xca)
+			if ((rsp->ccode == IPMI_CC_REQ_DATA_INV_LENGTH
+			     || rsp->ccode == IPMI_CC_REQ_DATA_FIELD_EXCEED
+			     || rsp->ccode == IPMI_CC_CANT_RET_NUM_REQ_BYTES)
```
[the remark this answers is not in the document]
pstrinkle: Roger.

lib/ipmi_fru.c
```diff
@@ -244,25 +244,25 @@ char *get_fru_area_str(uint8_t *data, uint32_t *offset)
  * input_filename - user input string
  *
  * returns   0  if path is ok
- * returns (-1) if path is NULL
- * returns (-2) if path is too short
- * returns (-3) if path is too long
+ * returns -1 if path is NULL
```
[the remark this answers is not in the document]
pstrinkle: So, I did each PR as a separate file(s) to allow for easier resolution of patch ordering, in this case, I'd prefer to just clean it up where I find it (in each PR).  I can go through what's missing after and clean it up as well.

lib/ipmi_fru.c
```diff
@@ -3857,34 +3857,26 @@ static int ipmi_fru_upg_ekeying(struct ipmi_intf *intf, char *pFileName,
 	    != 0) {
 		lprintf(LOG_ERR, "Failed to get multirec from file '%s'.",
 			pFileName);
-		if (buf) {
-			free(buf);
-			buf = NULL;
-		}
+		free(buf);
+		buf = NULL;
```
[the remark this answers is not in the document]
pstrinkle: Yeah, I wasn't sure how you felt about gotos.  Some people being so vehemently against them even though they can provide situations like this a cleaner exit path.

lib/ipmi_fru.c
```diff
@@ -1652,7 +1653,7 @@ typedef struct OemKontronInformationRecordV1 {
 static void ipmi_fru_oemkontron_get(int argc, char **argv, uint8_t *fru_data,
 				    int off, struct fru_multirec_oem_header *oh)
 {
-	static int badParams = FALSE;
```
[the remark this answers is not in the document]
pstrinkle: I have a task open to clean those up after these PRs -- sort of a "now that there's stdbool, fix it everywhere you find it and drop those macros"

pstrinkle:
Alexander, thanks for the comments.  I'll make the fixes in this PR where appropriate, however, I would prefer to keep the PRs to be one module each to simplify merging and dealing with conflicts.

pstrinkle:
Alexander, PTAL, everything you mentioned should now be fixed.

lib/ipmi_fru.c
```diff
@@ -600,16 +610,12 @@ int write_fru_area(struct ipmi_intf *intf, struct fru_info *fru, uint8_t id,
...
-		    || rsp->ccode == IPMI_CC_REQ_DATA_FIELD_EXCEED
-		    || rsp->ccode == IPMI_CC_CANT_RET_NUM_REQ_BYTES) {
-			if (fru->max_write_size > 8) {
-				fru->max_write_size -= 8;
-				lprintf(LOG_INFO,
-					"Retrying FRU write with request size %d",
-					fru->max_write_size);
-				continue;
-			}
+		if (fru_cc_rq2big(rsp->ccode) && fru->max_write_size > 8) {
```
AlexanderAmelkin: What is `8`, by the way?

lib/ipmi_fru.c
```diff
@@ -600,16 +610,12 @@ int write_fru_area(struct ipmi_intf *intf, struct fru_info *fru, uint8_t id,
...
+			fru->max_write_size -= 8;
+			lprintf(LOG_INFO,
+				"Retrying FRU write with request size %d",
+				fru->max_write_size);
+			continue;
 		} else if (rsp->ccode == 0x80) {
```
AlexanderAmelkin: I suppose this `0x80` must actually be `FRU_CC_WRITE_PROTECTED` (not yet defined anywhere, must go to ipmi_fru.h). There also must be `FRU_CC_DEVICE_BUSY` (0x81). Those are the two FRU command-specific completion codes.

lib/ipmi_fru.c
```diff
@@ -740,9 +746,7 @@ int read_fru_area(struct ipmi_intf *intf, struct fru_info *fru, uint8_t id,
 			/* if we get C7h or C8h or CAh return code then we
 			 * requested too many bytes at once so try again with
 			 * smaller size */
-			if ((rsp->ccode == IPMI_CC_REQ_DATA_INV_LENGTH
-			     || rsp->ccode == IPMI_CC_REQ_DATA_FIELD_EXCEED
-			     || rsp->ccode == IPMI_CC_CANT_RET_NUM_REQ_BYTES)
+			if (fru_cc_rq2big(rsp->ccode)
 			    && fru->max_read_size > 8) {
 				if (fru->max_read_size > 32) {
```
AlexanderAmelkin: I really wonder what those 8 and 32 are... I know they aren't yours. :)

lib/ipmi_fru.c
```diff
@@ -3849,40 +3853,41 @@ static int ipmi_fru_upg_ekeying(struct ipmi_intf *intf, char *pFileName,
...
 	}
 	if (write_fru_area(intf, &fruInfo, fruId, 0, offFruMultiRec,
 			   fileMultiRecSize, buf)
 	    != 0) {
 		lprintf(LOG_ERR, "Failed to write FRU area.");
+		ERR_EXIT;
+	}
+
+exit:
+	if (buf) {
```
AlexanderAmelkin: As you noted before, this check is superfluous. `free()` does nothing on NULL argument. That's standardized. It's safe to call `free(NULL)`. So let's put this `free()` right before `return rc` and get rid of this check and the `buf = NULL`.

lib/ipmi_fru.c
```diff
@@ -3849,40 +3853,41 @@ static int ipmi_fru_upg_ekeying(struct ipmi_intf *intf, char *pFileName,
...
 		free(buf);
 		buf = NULL;
-		return -1;
 	}
-	free(buf);
-	buf = NULL;
-	lprintf(LOG_INFO, "Done upgrading Ekey.");
-	return 0;
+
+	if (!rc) {
```
AlexanderAmelkin: Looks like it's safe to assume that anywhere above `exit:` but after the closing curly of the last `if()` is an `!rc` zone. So let's just move this message there and get rid of this check.

lib/ipmi_fru.c
```diff
@@ -90,6 +90,16 @@ static const char *chassis_type_desc[] = {"Unspecified",
 					  "Blade",
 					  "Blade Enclosure"};
 
+static inline bool fru_cc_rq2big(int code) {
+	if (code == IPMI_CC_REQ_DATA_INV_LENGTH
+	    || code == IPMI_CC_REQ_DATA_FIELD_EXCEED
+	    || code == IPMI_CC_CANT_RET_NUM_REQ_BYTES) {
+		return true;
```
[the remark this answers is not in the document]
pstrinkle: That code pattern isn't my first instinct due to a coding style guide during my formative years that precluded that.

lib/ipmi_fru.c
pstrinkle: I was going to put those in ipmi_cc.h so that all the return codes, including the command-specific ones could be found in one place, however, you disagree --> why not put them together?

lib/ipmi_fru.c
pstrinkle: It's now standardized, because for years and years it wasn't.  Are we sure someone compiling this isn't going to run into trouble with that?  In this case, they can reach buf with a null pointer as it's not a failure case in this specific method.

pstrinkle: Also, in a previous PR you indicated I needed to leave in the setting of NULL to a pointer, and this included cases where that had no effect, such as at the end of a method -- maybe you meant to exclude this case.

lib/ipmi_fru.c
pstrinkle: Presumably they're using 8 as the magic number for how many bytes to step down their write if it was rejected -- let's try 8 fewer bytes and try again repeatedly while we can.

I could name it, "FRU_WRITE_STEP"

pstrinkle:
Alexander, everything should be fixed except the null check ahead of free().  I talked to a few other engineers and there wasn't an indication of it being standardized.  We can address that kind of change in a separate PR once the other dominos fall.

AlexanderAmelkin:
Your fellow engineers are wrong in regard to null checking prior to free(). This is an excerpt from "Single UNIX Specification, System Interfaces and Headers, Issue 4, Version 2", as well as from "The Open Group Base Specifications Issue 7, 2018 edition" aka "IEEE Std 1003.1-2017 (Revision of IEEE Std 1003.1-2008)" aka "POSIX.1":

> **NAME**
>    _free_ — free allocated memory
> **SYNOPSIS**
> ```
> #include <stdlib.h>
> void free(void *ptr);
> ```
> **DESCRIPTION**
> The _free( )_ function causes the space pointed to by ptr to be deallocated; that is,
> made available for further allocation. **If ptr is a null pointer, no action occurs.**

It's been there in POSIX.1 for at least a decade, as far as I can recall.

pstrinkle:
Updated to use free_n().

lib/ipmi_fru.c
```diff
@@ -1520,7 +1520,7 @@ static void fru_area_print_multirec(struct ipmi_intf *intf,
  * returns : TRUE if data changed
  * returns : FALSE if data not changed
  */
-int ipmi_fru_query_new_value(uint8_t *data, int offset, size_t len)
+static int ipmi_fru_query_new_value(uint8_t *data, int offset, size_t len)
```
AlexanderAmelkin: `static` must be on a separate line. as well as `int`.
https://github.com/ipmitool/ipmitool/wiki/Coding-Standards#function-definition
That's been inherited from the old coding style used at SourceForge by this project.
As always, it hasn't been strictly followed, as you can see. But let's follow it at least where we change things.

include/ipmitool/ipmi_fru.h
```diff
@@ -127,13 +127,13 @@ struct fru_area_product {
...
-#define FRU_RECORD_TYPE_BASE_COMPATIBILITY 0x04
-#define FRU_RECORD_TYPE_EXTENDED_COMPATIBILITY 0x05
-#define FRU_RECORD_TYPE_OEM_EXTENSION	0xc0
+#define FRU_RECORD_TYPE_POWER_SUPPLY_INFORMATION	0x00
+#define FRU_RECORD_TYPE_DC_OUTPUT			0x01
+#define FRU_RECORD_TYPE_DC_LOAD				0x02
+#define FRU_RECORD_TYPE_MANAGEMENT_ACCESS		0x03
+#define FRU_RECORD_TYPE_BASE_COMPATIBILITY		0x04
+#define FRU_RECORD_TYPE_EXTENDED_COMPATIBILITY		0x05
+#define FRU_RECORD_TYPE_OEM_EXTENSION			0xc0
```
AlexanderAmelkin: Although I do like how it looks, I know that this will lead in future to further reformatting of the whole block, which will hinder git-blaming it. Hence, there is a rule to not use pretty formatting:
https://github.com/ipmitool/ipmitool/wiki/Coding-Standards#pretty-formatting

include/ipmitool/ipmi_fru.h
```diff
@@ -503,11 +503,11 @@ struct fru_picmgext_amc_p2p_record {
 	unsigned char resource_id         :4;
 	unsigned char /* reserved */      :3;
 	unsigned char record_type         :1;
-#else	
+#else
 	unsigned char record_type         :1;
 	unsigned char /* reserved */      :3;
 	unsigned char resource_id         :4;
-#endif 
+#endif
```
AlexanderAmelkin: Just a side note. We'll have to get rid of bit-fields in future. There is #59 for that.

include/ipmitool/ipmi_fru.h
```diff
@@ -503,11 +503,11 @@ struct fru_picmgext_amc_p2p_record {
 	unsigned char resource_id         :4;
 	unsigned char /* reserved */      :3;
 	unsigned char record_type         :1;
-#else	
+#else
 	unsigned char record_type         :1;
 	unsigned char /* reserved */      :3;
 	unsigned char resource_id         :4;
-#endif
```
AlexanderAmelkin: Thanks for cleaning up these hanging spaces.

lib/ipmi_fru.c
```diff
@@ -484,6 +484,9 @@ void free_fru_bloc(t_ipmi_fru_bloc *bloc)
 	}
 }
 
+/* By how many bytes to reduce a write command on a size failure. */
+#define FRU_AREA_IO_STEP	8
```
AlexanderAmelkin: As I said earlier, it's more like `FRU_BLOCK_SZ`.

lib/ipmi_fru.c
pstrinkle: I thought I had removed anything I added, as all free calls were changed to free_n()

include/ipmitool/ipmi_fru.h
pstrinkle: Clang-format handled that -- clang-format handles most of the cleanup in general, even if the end result needs some manual cleanup it's superior at whitespace adjustments.

pstrinkle:
It looks like everything is addressed, please let me know.

pstrinkle:
To make reviewing easier on you, and fixing on me, let's focus on this PR until it's ready and merged before moving onto the next one.

Regarding general cleanup:  I have no desire to hand fix an entire file, but I am completely willing to fix up after clang-format's mistakes.  Ideally, my changes to fix bugs could be more pointed, but the ever expansive impact of a change involves fixing all related lines, and I truly enjoy cleaning up code (using bools where possible, fixing other weirdness, etc) -- but doing whitespace manually for a 4000+ long file is really something that's easier cleaned up after an automated tool makes a best effort.  I have done experiments with setting clang-format to 200 or 300 columns so that it doesn't do anything odd with the functions or some of the very deep indentations.  Albeit, given time I imagine many of those deep indentations will be addressed in various ways.

pstrinkle:
Tried rebasing out the huge cleanup patch.  Lead to issues. Dropping this PR after cherry-picking the meat of it onto another branch.

lib/ipmi_fru.c
```diff
@@ -106,39 +106,70 @@ void ipmi_intf_set_max_response_data_size(struct ipmi_intf *intf,
...
-						   uint32_t *pRetLocation,
-						   uint32_t *pRetSize);
-static int ipmi_fru_get_multirec_from_file(char *pFileName, uint8_t *pBufArea,
-					   uint32_t size, uint32_t offset);
-static int ipmi_fru_get_multirec_size_from_file(char *pFileName,
-						uint32_t *pSize,
-						uint32_t *pOffset);
+static
+void
+ipmi_fru_read_to_bin(struct ipmi_intf *intf, char *pFileName,
```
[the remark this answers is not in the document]
pstrinkle: Yeah, flipped the words in my head.  Really the definitions and declarations are currently in a state violating the style guide, so both can be fixed.

PR #83, opened 2018-12-05
helper: add free_n method to handle clearing pointers

pstrinkle:
free_n() will free the memory and clear the pointer, which will reduce
the probability a developer will forget to clear the pointer after
freeing.
Resolves: #79


pstrinkle:
Fixed the extra merge patch by rebasing, now ready for review.

PR #86, opened 2018-12-21
ID:48 - Fixes for "Unable to send command: Invalid argument" on DELL …

ravisaripalli:
…server for

	    sensor who sensor owner ID is not 0x20.

	This fix is for sdr command error "Unable to send command: Invalid argument"
	on the dell servers for those sensor whoes sensor owner id is not 0x20.
	Adding a check to verfiy sensor owner id bit
	If the least significant bit of target address is set to 1
	Then no need for bridging the command since it is not an ipmb
	address

Committer: ravisaripalli <ravi_saripalli@dell.com>


ravisaripalli:
Hi Alexander 
I addressed the review comments, can you have a look.

PR #90, opened 2019-01-23
Fixed two issues found by "cppcheck".

jmoellers:
Fixed a resource leak by closing an open file before returning.
Removed an "if" with a condition which is always true to avoid confusion.
Also removed a call to "time()" in order to have two timestamps identical
at all times.

### DIFF
--- a/lib/ipmi_hpmfwupg.c
+++ b/lib/ipmi_hpmfwupg.c
@@ -1398,6 +1398,7 @@ HpmfwupgGetBufferFromFile(char *imageFilename,
 	if (ret != 0) {
 		lprintf(LOG_ERR, "Failed to seek in the image file '%s'",
 				imageFilename);
+		fclose(pImageFile);
 		return HPMFWUPG_ERROR;
 	}
 	pFwupgCtx->imageSize  = ftell(pImageFile);
@@ -2290,13 +2291,9 @@ HpmfwupgWaitLongDurationCmd(struct ipmi_intf *intf,
 			}
 		}
 	}
-	if (rc == HPMFWUPG_SUCCESS) {
-		/* Poll upgrade status until completion or timeout*/
-		timeoutSec1 = time(NULL);
-		timeoutSec2 = time(NULL);
-		rc = HpmfwupgGetUpgradeStatus(intf, &upgStatusCmd,
-				pFwupgCtx, 1);
-	}
+	/* Poll upgrade status until completion or timeout*/
+	timeoutSec2 = timeoutSec1 = time(NULL);
+	rc = HpmfwupgGetUpgradeStatus(intf, &upgStatusCmd, pFwupgCtx, 1);
 	while (
 			/* With KCS: Cover the case where we sometime
 			 * receive d5 (on the first get status) from


lib/ipmi_hpmfwupg.c
```diff
@@ -1398,6 +1398,7 @@ HpmfwupgGetBufferFromFile(char *imageFilename,
 	if (ret != 0) {
 		lprintf(LOG_ERR, "Failed to seek in the image file '%s'",
 				imageFilename);
+		fclose(pImageFile);
```
AlexanderAmelkin: I'd prefer a Linux kernel style [centralized exiting](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#centralized-exiting-of-functions).
Also, I'd set `rc` to `HPMFWUPG_ERROR` by default and only set it to `HPMFWUPG_SUCCESS` when we're right before the end of the successful execution path. That is usually right before that centralized exit point.

lib/ipmi_hpmfwupg.c
```diff
@@ -2290,13 +2291,9 @@ HpmfwupgWaitLongDurationCmd(struct ipmi_intf *intf,
...
-	if (rc == HPMFWUPG_SUCCESS) {
-		/* Poll upgrade status until completion or timeout*/
-		timeoutSec1 = time(NULL);
-		timeoutSec2 = time(NULL);
-		rc = HpmfwupgGetUpgradeStatus(intf, &upgStatusCmd,
-				pFwupgCtx, 1);
-	}
+	/* Poll upgrade status until completion or timeout*/
+	timeoutSec2 = timeoutSec1 = time(NULL);
+	rc = HpmfwupgGetUpgradeStatus(intf, &upgStatusCmd, pFwupgCtx, 1);
```
AlexanderAmelkin: I totally support this change. Thank you.

jmoellers:
On 23.01.19 17:55, Alexander Amelkin wrote:
> *@AlexanderAmelkin* requested changes on this pull request.
> 
> Please prepend the commit summary with 'hpm: ', and also please make two
> commits instead of one. One per change. The changes are different in
> nature so it would be good to have them committed separately (would be
> easier to revert in the event of any problems).

Will do that.

> 
> Thank you for your help.
> 
> ------------------------------------------------------------------------
> 
> In lib/ipmi_hpmfwupg.c
> <https://github.com/ipmitool/ipmitool/pull/90#discussion_r250280146>:
> 
>> @@ -1398,6 +1398,7 @@ HpmfwupgGetBufferFromFile(char *imageFilename,
>  	if (ret != 0) {
>  		lprintf(LOG_ERR, "Failed to seek in the image file '%s'",
>  				imageFilename);
> +		fclose(pImageFile);
> 
> I'd prefer a Linux kernel style centralized exiting

There are 4 "return"s in that function, 3 of which should have to go.
It'd be a pretty big rewrite if I'd change that to a centralized return,
possibly ignoring your request above ;-)
So I'll just fix this bug and do a rewrite of that function in a second
pull request.

> <https://www.kernel.org/doc/html/v4.10/process/coding-style.html#centralized-exiting-of-functions>.
> Also, I'd set |rc| to |HPMFWUPG_ERROR| by default and only set it to
> |HPMFWUPG_SUCCESS| when we're right before the end of the successful
> execution path. That is usually right before that centralized exit point.


> 
> ------------------------------------------------------------------------
> 
> In lib/ipmi_hpmfwupg.c
> <https://github.com/ipmitool/ipmitool/pull/90#discussion_r250280589>:
> 
>> @@ -2290,13 +2291,9 @@ HpmfwupgWaitLongDurationCmd(struct ipmi_intf *intf,
>  			}
>  		}
>  	}
> -	if (rc == HPMFWUPG_SUCCESS) {
> -		/* Poll upgrade status until completion or timeout*/
> -		timeoutSec1 = time(NULL);
> -		timeoutSec2 = time(NULL);
> -		rc = HpmfwupgGetUpgradeStatus(intf, &upgStatusCmd,
> -				pFwupgCtx, 1);
> -	}
> +	/* Poll upgrade status until completion or timeout*/
> +	timeoutSec2 = timeoutSec1 = time(NULL);
> +	rc = HpmfwupgGetUpgradeStatus(intf, &upgStatusCmd, pFwupgCtx, 1);
> 
> I totally support this change. Thank you.

'welcome.

Josef

PR #99, opened 2019-02-25
Cleanup/nm

AlexanderAmelkin:
An effort to separate the `nm` module and get rid of magic numbers in it.

@pstrinkle, @vmauery, can you please check that these changes don't break the `nm` functionality? I don't have any NM-enabled hardware to test against.


AlexanderAmelkin:
Apparently, the attempt to properly check completion code dictionaries failed because of wrong assumption regarding behavior of `val2str`. Will re-fix that part.

pstrinkle:
I don't have an nm enabled system.  I checked this morning and none of the commands returned anything but 0xc1.

AlexanderAmelkin:
> I don't have an nm enabled system. I checked this morning and none of the commands returned anything but 0xc1.

Well, that's expected for non-nm systems.

pstrinkle:
> > I don't have an nm enabled system. I checked this morning and none of the commands returned anything but 0xc1.
> 
> Well, that's expected for non-nm systems.

Of course -- I actually didn't know what to expect for the commands -- I didn't know if my system was nm-enabled.  I'd never looked at that module, so it's possible there was some default behavior in openbmc that I inherited.  Good to know there isn't.  May be something to add to OpenBMC at some point, if someone isn't already pushing that.

include/ipmitool/ipmi_nm.h
```diff
@@ -169,3 +173,13 @@ struct nm_suspend {
...
 int ipmi_nm_main(struct ipmi_intf *intf, int argc, char **argv);
+
+static inline void nm_set_id(void *buf)
+{
+	htoipmi24(IPMI_NM_ID, buf);
+}
+
+static inline bool nm_check_id(void *buf)
+{
+	return IPMI_NM_ID != ipmi24toh(buf);
```
vmauery: this should be:
	return IPMI_NM_ID == ipmi24toh(buf);

Because when it is called, it bails if this function returns false.

vmauery:
After fixing the nm_check_id function, the nm commands work about like they did before.

lib/ipmi_dcmi.c
```diff
@@ -2056,1485 +1766,123 @@ ipmi_dcmi_pwr_actdeact(struct ipmi_intf * intf, uint8_t option)
...
-	req.msg.data = msg_data;
-	req.msg.data_len = 5;
-	rsp = intf->sendrecv(intf, &req);
-	if (chk_nm_rsp(rsp)) {
-		return -1;
-	}
-	memcpy(caps, rsp->data, sizeof (struct nm_capability));
-	return 0;
-}
+	if ( argc == 10) {
```
pstrinkle: nit: extra whitespace

pstrinkle: I see now this is just moved code, so perhaps it should be left as-is in that case.

lib/ipmi_dcmi.c
```diff
@@ -3552,8 +1900,8 @@ ipmi_dcmi_parse_power(struct ipmi_intf * intf, int argc, char **argv)
 	case 0x02:
 		/* set limit */
 		if (argc < 4) {
-			print_strs(dcmi_pwrmgmt_set_usage_vals,
-			           "set_limit <parameter> <value>",
+			dcmi_print_strs(dcmi_pwrmgmt_set_usage_vals,
+			                "set_limit <parameter> <value>",
```
pstrinkle: you fixed one indentation change for the longer function name, but not the next one.  could be argued either way that it should also be changed.

lib/ipmi_dcmi.c
```diff
@@ -3639,9 +1987,11 @@ ipmi_dcmi_thermalpolicy(struct ipmi_intf * intf, int argc, char **argv)
 			        argv[3]);
 			return (-1);
 		}
-		persistanceFlag = (uint8_t) str2val2(argv[4], dcmi_thermalpolicy_set_parameters_vals);
-		actionHardPowerOff = (uint8_t) str2val2(argv[5], dcmi_thermalpolicy_set_parameters_vals);
-		actionLogToSEL = (uint8_t) str2val2(argv[6], dcmi_thermalpolicy_set_parameters_vals);
+
+		persistanceFlag =(uint8_t)dcmi_str2val(argv[4], dcmi_thermalpolicy_set_parameters_vals);
```
pstrinkle: nit: you accidentally deleted the extra space this had around the assignment.

lib/ipmi_nm.c
```diff
@@ -0,0 +1,1834 @@
...
+	req.msg.data = msg_data;
+	req.msg.data_len = 5;
+	rsp = intf->sendrecv(intf, &req);
+	if (is_nm_error(rsp)) {
+		return -1;
+	}
+	memcpy(policy, rsp->data, sizeof(struct nm_get_policy));
+	return 0;
+}
+static int
```
pstrinkle: I'm assuming since this is just moved code, there'll be possibly follow-on for style conformance?

vmauery:
Just for the record, in order to test this, I had to bridge to the ME, like this:
# ipmitool -b 6 -t 0x2c nm statistics power
    Power domain:                             platform
    Policy/Global Admin state                 Globally Enabled
    Policy/Global Operational state           suspended
    Policy/Global Measurement state           in progress
    Policy Activation state                   not triggered
    Instantaneous reading:                         156 Watts
    Minimum during sampling period:                130 Watts
    Maximum during sampling period:                202 Watts
    Average reading over sample period:            153 Watts
    IPMI timestamp:                           Thu Feb 28 18:51:42 2019
    Sampling period:                          00001420 Seconds.

Trying to send the nm commands to the BMC itself will give C1 because it is the ME that needs to respond, not the BMC. Your bridge and address for the ME may be different. Consult your board manufacturer.

include/ipmitool/ipmi_nm.h
AlexanderAmelkin: Great catch! Will fix. Thanks.

Issue #11, opened 2018-05-30
PEN file can't be generated when direct internet connection isn't available


AlexanderAmelkin:
In #6 by @sll552 wrote:

> I had problems building the source in an environment without direct internet access. It seems that the generated pen file is not properly clean when using make clean and an invalid file is generated when no direct internet connection is available, so curl fails.

It looks like an improvement or a workaround/documentation is needed for not-directly-connected build machines. Patches are welcome.

AlexanderAmelkin:
The missing curl problem should be addressed in context of this issue as well.

AlexanderAmelkin:
Ideally, build dependency on `curl` and `iconv` must be eliminated. Instead, `ipmitool` needs to be taught to read the IANA PEN file from local filesystem (e.g. from `/usr/share/misc/ipmi.ids`) like `lspci` and `lsusb` do.

AlexanderAmelkin:
Mistakenly auto-closed by github upon seeing a reference in commit log. Reopening.

PR #105, opened 2019-02-27
create_pen_list: only print if values are set

vmauery:
If a build environment doesn't have network access the fetch of the IANA PEN list will fail and the create_pen_list script will create invalid output causing the build to fail. This fixes the parsing.

### DIFF
--- a/lib/create_pen_list
+++ b/lib/create_pen_list
@@ -67,7 +67,9 @@ parse_pen_list() {
       }
 
       END {
-          print "{ " PEN ", \"" ENTERPRISE "\" },"
+          if(PEN) {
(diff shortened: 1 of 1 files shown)


AlexanderAmelkin:
Great. I'm willing to merge this for now, but it looks like it is related to #11, isn't it? If so, please put a "Partially resolves ipmitool/ipmitool#11" into the commit message.

vmauery:
> can it work,when the server is offline?

This change only affects the building of ipmitool. If the remote server that the PEN list is downloaded from is unavailable, or the machine building ipmitool is offline, then this allows the build to succeed, using a mostly-empty list instead of the full IANA PEN list.

Issue #46, opened 2018-10-30
exchange-bmc-os-info: Remove dependency on ipmi.service


charles-rose:
Hello,
The current exchange-bmc-os-info systemd unit waits for ipmi.service to start. This is provided by OpenIPMI and is not necessary as the kernel modules load automatically during system bootup[1]. A better method is to look for the presence of the KCS device node - /dev/ipmi* which is created before this systemd units are run.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/char/ipmi/ipmi_devintf.c?h=v4.19&id=6c89161b10f5771ee0b51ada0fce0e8835e72ade

charles-rose:
Created https://github.com/ipmitool/ipmitool/pull/45

PR #115, opened 2019-04-09
exchange-bmc-os-info: Remove dependency on ipmi.service

charles-rose:
Resolves ipmitool/ipmitool#46

Most modern systems do not require ipmi.service to load the kernel
modules. Checking for /dev/ipmi* would be sufficient.

v2: Use Assert in place of Condition to explicitly fail based on
   AlexanderAmelkin's feedback.
Signed-off-by: Charles Rose <charles.rose@dell.com>


AlexanderAmelkin:
Thank you. Next time please don't open a new PR. Just force-push to the old branch to update the existing PR.

PR #117, opened 2019-04-15
Add mechanism to configure to set the default interface

vmauery:
In some cases, the user may want to have a different default interface
without the need to always specify it on the command line. Add a
configure option that sets the default interface without the need to
patch the code.

Configure as: ./configure DEFAULT_INTF=name
where name is an interface name that might be enabled with
--enable-intf-<name>.

The configure will enforce that the selected default interface is
enabled.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -671,6 +671,25 @@ AC_TRY_COMPILE([],[
 		[Define to 1 if you need to use #pragma pack instead of __attribute__ ((packed))])]
 )
 
+dnl if no environment variable is set, set the default value for the default intf
+if test "${xenable_intf_open}" = "xyes"; then
+   DEFAULT_INTF_NO_ENV=open
+else dnl macOS does not build open interface, it defaults to lan
+   DEFAULT_INTF_NO_ENV=lan
+fi
+dnl allow for a default interface to be set on configure
+AC_ARG_VAR(DEFAULT_INTF, [Set the default interface to use (default=${DEFAULT_INTF_NO_ENV})])
+
+dnl set the default value for the default interface environment variable
+if test "x${DEFAULT_INTF}" == "x"; then
+	echo "DEFAULT_INTF not found in environment; setting to ${DEFAULT_INTF_NO_ENV}"
+	DEFAULT_INTF=${DEFAULT_INTF_NO_ENV}
+fi
+
+xdefault_intf_is_enabled="xenable_intf_${DEFAULT_INTF}"
+if test "x${!xdefault_intf_is_enabled}" != "xyes"; then
+	AC_MSG_ERROR([** Cannot set ${DEFAULT_INTF} as default; ${DEFAULT_INTF} is not enabled. :${!xdefault_intf_is_enabled}:])
+fi
 
 dnl Generate files for build
 AC_CONFIG_FILES([Makefile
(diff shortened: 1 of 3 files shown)


src/plugins/ipmi_intf.c
```diff
@@ -128,17 +128,40 @@ struct ipmi_intf * ipmi_intf_table[] = {
...
+	 * to return some valid value here, so the first entry works
+	 */
+	return ipmi_intf_table[0];
+}
+
 /* ipmi_intf_print  -  Print list of interfaces
  *
  * no meaningful return code
  */
 void ipmi_intf_print(struct ipmi_intf_support * intflist)
```
AlexanderAmelkin: Funny thing is that this `intflist` is never actually used and there is no hint as to why it has been introduced in the first place back in 2005 by Duncan...

src/plugins/ipmi_intf.c
```diff
@@ -128,17 +128,40 @@ struct ipmi_intf * ipmi_intf_table[] = {
...
+ *
+ * returns a valid interface pointer
+ */
+static struct ipmi_intf *get_default_interface(void)
+{
+	static const char *default_intf_name = DEFAULT_INTF;
+	struct ipmi_intf ** intf;
+	for (intf = ipmi_intf_table; intf && *intf; intf++) {
+		if (strncmp(default_intf_name, (*intf)->name,
+		            strlen(default_intf_name)) == 0)
```
AlexanderAmelkin: Missed this. What's the purpose of `strncmp` here? Why not just `strcmp` ? I've been lately eradicating these weird comparisons everywhere throughout the code. If you were just mimicking the habits of previous contributors, please don't. And please don't compare `strXcmp` to 0 explicitly. This is idiomatic enough:

`if (!strcmp(default_intf_name, (*intf)->name)) {`

configure.ac
```diff
@@ -671,6 +682,10 @@ AC_TRY_COMPILE([],[
 		[Define to 1 if you need to use #pragma pack instead of __attribute__ ((packed))])]
 )
 
+xdefault_intf_is_enabled="xenable_intf_${DEFAULT_INTF}"
+if test "x${!xdefault_intf_is_enabled}" != "xyes"; then
+	AC_MSG_ERROR([** Cannot set ${DEFAULT_INTF} as default; ${DEFAULT_INTF} is not enabled. :${!xdefault_intf_is_enabled}:])
```
AlexanderAmelkin: Apparently, this fails the build for MacOS where the default 'open' interface isn't enabled. I honestly don't know what is the default interface for Mac or why 'open' isn't enabled there (probably, some libs are missing). Nonetheless without your changes it was building fine somehow. This build failure must be fixed.

src/plugins/ipmi_intf.c
vmauery: It was a cut and paste from below. Fixed.

configure.ac
vmauery: It is funny that the part that I added to make it work more intuitively broke things. I added this part to be able to print out the default value of the default interface variable. And on Linux, that is open. I added a check to see if open is not available (or not building) and default to lan, which is likely universally built in that case. I tested it by forcing a build that disables open.

src/plugins/ipmi_intf.c
vmauery: I noticed that too. Maybe it was part of a half-baked dynamically loaded interface mechanism.

Issue #118, opened 2019-04-17
encounter Segmentation fault  when using ipmitool


leye7755:
Core was generated by` ipmitool -I lanplus -H xxxxxx -U xxxx -P XXXXXXXXX raw 0x3a 0x71 0x01`
Program terminated with signal 11, Segmentation fault.

Why does this problem arise?  We use popen to use ipmitool

leye7755:
@AlexanderAmelkin

leye7755:
Looking forward to and thanking you for your help。@AlexanderAmelkin

leye7755:
using sdr module  encountered a problem:
debug details:
ipmi_sdr.c : ipmi_ipmi_sdr_read_sensor_value() call ipmi_sdr_get_unit_string( intf, pct, type, base,modifier)
and  base is 255 which exceed the size of unit_desc array  .   leading to Segmentation fault。

AlexanderAmelkin:
Can you please run `ipmitool` with nine `-v` arguments (that is, `ipmitool -vvvvvvvvv ...`) and attach (don't copy-paste) the output here? It's unclear why `base` is 255. Looks like an invalid response from your BMC. I'd like to see the full communication log to understand how to properly handle it.

leye7755:
@AlexanderAmelkin Can i send # your email? can't find the way to attach log.

AlexanderAmelkin:
Please check the branch from PR #120

AlexanderAmelkin:
As for attachments, there is the "Attach files by dragging..." string right below the comment entry box here on this page.

AlexanderAmelkin:
Sorry, committed the wrong version there. Will fix soon.

leye7755:
[sdr.log](https://github.com/ipmitool/ipmitool/files/3101861/sdr.log)
this is sdr log

AlexanderAmelkin:
@leye7755, I don't see a segmentation fault in the attached log. Is it the log from #120 ?

leye7755:
[sdr-20190423.log](https://github.com/ipmitool/ipmitool/files/3105483/sdr-20190423.log)
I add comment  in ipmi_sdr.c : ipmi_sdr_get_unit_string() 
printf("#test:%d#\n",base);

AlexanderAmelkin:
I now see it hit the wrong string for `CPU1 DTS` sensor.
That's quite expected with the old code, there is no range check in it.
Have you tried running the code from #120 ?

[...]

PR #120, opened 2019-04-19
sdr: Fix segfault on invalid unit types

AlexanderAmelkin:
The program would crash if the BMC returned an out of range (>90)
unit type for a full sensor record. This commit adds a range check
and also add support for IPMI 2.0 additional unit types 91 and 92
("fatal error" and "grams").

Resolves ipmitool/ipmitool#118

Signed-off-by: Alexander Amelkin <alexander@amelkin.msk.ru>

### DIFF
--- a/include/ipmitool/ipmi_sdr.h
+++ b/include/ipmitool/ipmi_sdr.h
@@ -37,6 +37,7 @@
 # include <config.h>
 #endif
 
+#include <stdbool.h>
 #include <inttypes.h>
 #include <math.h>
 #include <ipmitool/bswap.h>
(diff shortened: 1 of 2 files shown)


leye7755:
#118 At present, I simply set the value beyond this range to 0.

AlexanderAmelkin:
@leye7755, does THIS patch work for you? Please build from this branch and tell me if it works:
```
# (assuming you're inside the cloned repo)
git checkout bugfix/118-fix-sdr-unit-segfault
./bootstrap
./configure
make
```

PR #125, opened 2019-05-20
Ipmitool sends incorrect format of platform event through SSIF. Generator id field is missing in the request data.

adarshgrunni:
Adding a proper condition for ssif in platform events to generate the generator id.

### DIFF
--- a/lib/ipmi_event.c
+++ b/lib/ipmi_event.c
@@ -94,7 +94,7 @@ ipmi_send_platform_event(struct ipmi_intf * intf, struct platform_event_msg * em
 	req.msg.data = rqdata;
 
 	chmed = ipmi_current_channel_medium(intf);
-	if (chmed == IPMI_CHANNEL_MEDIUM_SYSTEM) {
+	if (chmed == IPMI_CHANNEL_MEDIUM_SYSTEM || chmed == IPMI_CHANNEL_MEDIUM_SMBUS_1 || chmed == IPMI_CHANNEL_MEDIUM_SMBUS_2) {
 		/* system interface, need extra generator ID */
 		req.msg.data_len = 8;
 		rqdata[0] = 0x41;   // As per Fig. 29-2 and Table 5-4
@@ -507,7 +507,7 @@ ipmi_event_fromfile(struct ipmi_intf * intf, char * file)
 	req.msg.data_len = 7;
 
 	chmed = ipmi_current_channel_medium(intf);
-	if (chmed == IPMI_CHANNEL_MEDIUM_SYSTEM) {
+	if (chmed == IPMI_CHANNEL_MEDIUM_SYSTEM || chmed == IPMI_CHANNEL_MEDIUM_SMBUS_1 || chmed == IPMI_CHANNEL_MEDIUM_SMBUS_2) {
 		/* system interface, need extra generator ID */
 		rqdata[0] = 0x41;   // As per Fig. 29-2 and Table 5-4
 		req.msg.data_len = 8;


AlexanderAmelkin:
Could you please elaborate on why this new condition is "proper" ? References (in the commit log message) to relevant sections of IPMI specification would be helpful.

Also, please see https://github.com/ipmitool/ipmitool/wiki#how-to-contribute for guidelines on commit messages.

AlexanderAmelkin:
References to IPMI spec in the comments above the conditional operator would be even better.

adarshgrunni:
image

As per IPMI spec SSIF uses the same format as the KCS .So when we send platform event messages through ssif the proper generator id should be filled by ipmitool . 
image

adarshgrunni:
In BMC side we are expecting an eight byte of data including generator id for platform event through SSIF. But the ipmitool  sends 7 bytes of date . So the bmc will consider the first byte as a generator id (Eventhough ipmi does't send the generator id ) . So entire event date will become in correct.

For example :
if user issue ipmitool event 1 through ssif 
we are expecting an sel event as :
01 | 05/29/2019 | 06:21:57 | Temperature #0x30 | Upper Critical going high | Asserted
But instead user will get : 
02 | 05/29/2019 | 06:25:10 | Unknown #0x01 | | Asserted

AlexanderAmelkin:
Thank you for the explanation. However you haven't updated the commit log. I did that for you by putting your explanation into the log (it still lacks your 'Signed-off-by').

However, now with your explanation I'm even more cautious about merging your change as it doesn't seem to match your explanation. The old code pretty well matches the IPMI spec in only adding the generator ID to the event message if the interface is a system interface. What you do in your patch is adding the generator ID for SMBUS (IPMB) interfaces, which must not be done as per Table 29-4 of IPMI spec:

image

As you see, only "System interface" has generator ID, and system interface is medium type 0x0C as per Table 6-3:

image

It is unclear why your BMC uses medium type 0x08 or 0x09 for a channel through which it accepts event messages. SMBUS media type is for channels to which sensors are connected (for use with Master Write/Read command). You must be using 0x0C for media type. I suspect you may have a problem in your BMC implementation and so far I am inclined towards rejecting this PR unless you have motivated objections.

Thank you.

adarshgrunni:
Hi Alex,

In ARM platforms KCS is not supported ,So instead of that we will use SSIF as a system interface .

As per IPMI Spec , we need to use channel medium type 0x08 or 0x09 for SSIF interface . If we change that to 0x0c it will be a violation of IPMI Spec .  Also in Event message section in IPMI spec says **"SSIF uses the same format as KCS"** so if we get event message through SSIF .Generator ID should be generated by ipmitool .

Thanks ,
Adarsh

AlexanderAmelkin:
Adarsh, I may be missing something, but I don't see where IPMI spec demands SSIF interface channel to be of SMBUS type and not of System Interface type. On the contrary, I see for example a 'Get System Interface Capabilities' command implying that it must be called to find out the physical medium of the channel of 'System Interface' type, and that commands can report that the system interface is either SSIF, KCS or SMIC.

I totally agree that for system interfaces (such as SSIF) there must be a 'Generator ID' present in the event message. However, I'm afraid that is not what you're doing with your patch. To me it looks like you're breaking some non-SSIF SMBus communication instead. Again, I believe the medium type for your SSIF must be 0x0C, not 0x08 or 0x09.

Nonetheless, I well may be wrong, so if you know a section in IPMI spec that clearly states that SSIF channel must have an 'SMBus' media type, please point me to it. I failed to find it.

adarshgrunni:
Hi Alex,
In the IPMI Spec Channel medium type they are mentioned system interface 0x0c only for KCS,SMIC and BT. Since SSIF(Smbus system interface) is not mentioned in system interface 0x0c ,we are using 0x9 Smbus as the channel medium type.  

Either SSIF should be included as a system interface in IPMI spec channel medium type 0xc. Other ways we should go with channel medium type as a Smbus .

Thanks,
Adarsh

AlexanderAmelkin:
Adarsh, in my opinion this is a mere lapse in the specification. They should have left that list open in Table 6-3. Unfortunately, there are lots of such ambiguities in this spec. They have section 1.7.16 "System Interfaces" that clearly states that there are FOUR system interfaces (KCS, SMIC, BT and SSIF), but then they have section 1.7.31 saying that "It is mandatory to implement a system interface that is compatible with one of the **three** specified system interfaces" without specifying which three of the four interfaces are meant. Then in section 6 "IPMI Messaging interfaces" they again say that "As mentioned earlier, there are three System Interface implementations specified for the BMC: SMIC, KCS, and BT". Is all looks like during update from 1.5 to 2.0 they have updated section 1.7.16, but forgot to update Table 6-3, section 1.7.31 and section 6. Yet again, there is 'Get System Interface Capabilities' command that has a parameter 'System Interface Type' that can specify that SI is of SSIF type.

You say that since Table 6-3 only lists 'KCS, SMIC or BT' for System Interface, then no other types are allowed. My point is that Table 6-3 is minor relative to Section 1.7.16 "System Interfaces", which lists SSIF as one of the possible media types.

Anyway, Table 6-1 says that system interface must be Channel number 0xF. Is it so for you? If it is, and it must be so as per section 6.3, and if you insist on using SMBUS media type, then ipmitool could be changed to check the channel number along with media type, and if the channel number is 0xF, then treat it the same way as if media type was 0xC.

I have prepared my version of the fix. Please see #138. I also did some code cleanup there. I still believe that your implementation is wrong and code 0x8 and 0x9 are for real SMBUS channels, not for SSIF. Nonetheless, the code from that PR should work with your implementation properly, unless you're also not using channel number 0xF for your SSIF.

PR #129, opened 2019-06-04
Feature/add support for boot mailbox

AlexanderAmelkin:
This is an effort to add support for the Boot Initiator Mailbox system boot option.
Before this change the option could only be dumped as a whole in hex dump mode using `bootparam get 7`. With this change set it becomes possible to dump a single block using `bootparam get 7 <id>` as well as using `bootmbox get block <id>`. Also added are text dump function (`bootmbox get text`), and the function to set the mailbox value in both text and binary modes (`bootmbox set [text] [block <id>]`).

### DIFF
--- a/doc/ipmitool.1
+++ b/doc/ipmitool.1
@@ -473,6 +473,34 @@ Force boot into BIOS Setup.
 
 Force boot from Floppy/primary removable media.
 .RE
+
+.TP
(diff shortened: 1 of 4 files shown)


AlexanderAmelkin:
@vmauery, @pstrinkle, I'd like you to review this change set if you have time.

Issue #128, opened 2019-06-02
ipmitool shouldn't default to lan intf


wkennington:
Without any configuration, ipmitool is now defaulting the lan interface. This is surprising to and end user who is used to being able to do `ipmitool <cmd>` on a machine that has a local ipmi interface. It's not particularly useful to default to the lan interface anyway since it requires additional, explicit configuration.

AlexanderAmelkin:
I believe this is because of #117, but it's quite surprising because as you may see the change introduced there takes care of defaulting to 'open' on Linux. It indeed defaults to 'lan' on MacOS. If that is wrong (sorry, I don't have MacOS), then please tell what must be the default interface for it or create a pull request yourself to refix #117.

wkennington:
I'd have to dig into it more but it is on linux, and the open interface was detected correctly.

[build.log](https://github.com/ipmitool/ipmitool/files/3249927/build.log)

PR #132, opened 2019-06-06
Fix default interface to behave as it did before

vmauery:
Prior to 95038ba01b99153de870dde91406bdc1c7265217 the default interface
was 'open' for builds where it was available and 'lan' otherwise. Due to
a logic error in 95038ba01b99153de870dde91406bdc1c7265217, the default
interface was always getting set to lan unless it was specified by the
environment variable. This commit fixes the logic and sets the default
to open if it is available and lan otherwise.

Resolves ipmitool/ipmitool#128

Tested: ./configure # no options reports 'open' as default
        ./configure --enable-intf-open=no # reports lan as default
	./configure --enable-intf-dbus DEFAULT_INTF=dbus
	# reports dbus as default

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -672,13 +672,13 @@ AC_TRY_COMPILE([],[
 )
 
 dnl if no environment variable is set, set the default value for the default intf
-if test "${xenable_intf_open}" = "xyes"; then
+if test "${xenable_intf_open}" = "yes"; then
    DEFAULT_INTF_NO_ENV=open
 else dnl macOS does not build open interface, it defaults to lan
    DEFAULT_INTF_NO_ENV=lan
 fi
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -672,13 +672,13 @@ AC_TRY_COMPILE([],[
 )
 
 dnl if no environment variable is set, set the default value for the default intf
-if test "${xenable_intf_open}" = "xyes"; then
```
AlexanderAmelkin: I knew there was something wrong with this "xyes"... Good things don't look like [that](https://translate.google.com/#view=home&op=translate&sl=ru&tl=en&text=%D1%85%D1%83%D0%B5%D0%B2). ;) Still I missed it. Mea culpa. Good to have that fixed now.

PR #135, opened 2019-06-13
Bugfix/11: Read IANA registry from file

### LINKED ISSUE #11: PEN file can't be generated when direct internet connection isn't available

AlexanderAmelkin:
This fixes #11. Need a review.


AlexanderAmelkin:
Ready for review

lib/ipmi_strings.c
```diff
@@ -754,3 +769,241 @@ const struct oemvalstr picmg_busres_shmc_status_vals[] = {
...
+		return 0;
+	}
+
+	/*
+	 * First try to open user's local registry if HOME is set
+	 */
+	if ((home = getenv("HOME"))) {
+		char path[PATH_MAX];
+		strncpy(path, home, sizeof(path));
+		strcat(path, "/.local" IANA_PEN_REGISTRY);
```
pstrinkle: Is there not a safer strcat?

lib/ipmi_strings.c
```diff
@@ -830,15 +830,15 @@ int ipmi_oem_info_init()
 	if ((home = getenv("HOME"))) {
 		char path[PATH_MAX];
 		strncpy(path, home, sizeof(path));
-		strcat(path, "/.local" IANA_PEN_REGISTRY);
+		strcat(path, "/" IANAUSERDIR "/" IANA_PEN_REGISTRY);
```
pstrinkle: similarly, is there not a safe strcat?

lib/ipmi_strings.c
```diff
@@ -830,15 +830,15 @@ int ipmi_oem_info_init()
...
 		in = fopen(path, "r");
 	}
 
 	if (!in) {
 		/*
 		 * Now open the system default file
 		 */
-		in = fopen(IANA_PEN_REGISTRY, "r");
+		in = fopen(IANADIR "/" IANA_PEN_REGISTRY, "r");
```
pstrinkle: I was of the impression some users were running this on windows?  Perhaps just with cygwin.

AlexanderAmelkin: Yes, so far ipmitool doesn't build with mingw32, only with cygwin.
I personally haven't built it (or anything else) for windows even with cygwin, so I just don't know how cygwin treats paths.

Anyway, if you know a portable way of building this path, I'd be glad to see it. I'll try to investigate that myself too...

lib/ipmi_strings.c
AlexanderAmelkin: Yes, should probably use `strncat()` instead. Or did you mean anything else?

pstrinkle: That's what I meant, thanks! (I hadn't looked it up just assumed there was an "n" variation).

AlexanderAmelkin: There is also `snprintf()` and some people say it is safer than `strncat()` but I don't see why.
Nonetheless, I'll probably use `snprintf()`. Hopefully that will make Codacy CI plugin happy.

lib/ipmi_strings.c
pstrinkle: I don't build for windows myself.  I found this:

const char kPathSeparator =
#ifdef _WIN32
                            '\\';
#else
                            '/';
#endif

https://stackoverflow.com/questions/12971499/how-to-get-the-file-separator-symbol-in-standard-c-c-or

Honestly, it's not something you need to fix, but it's something I wanted to ask about as I know some people compile this.  If mingw32 works with "/" fine then this is fine.

AlexanderAmelkin: I believe I found the way. See the updated patch set.

lib/ipmi_strings.c
```diff
@@ -754,3 +769,251 @@ const struct oemvalstr picmg_busres_shmc_status_vals[] = {
...
+	}
+
+	if (!in) {
+		/*
+		 * Now open the system default file
+		 */
+		in = fopen(IANADIR PATH_SEPARATOR IANA_PEN_REGISTRY, "r");
+		if (!in) {
+			lperror(LOG_ERR, "IANA PEN registry open failed");
+			exit(-1); /* Can't leave ipmi_oem_info uninitialized */
```
vmauery: Rather than having the process exit if a PEN list cannot be found, why not just run with an empty one? Sure there will not be complete data, but this seems like a bit of a draconian measure. ipmitool may be wounded without this, but I don't think it is a fatal wound.

AlexanderAmelkin: I agree. Will change it to use an empty dummy array if allocation fails.

AlexanderAmelkin:
Also ran through valgrind, found a couple of memory allocation bugs, fixed them.

Issue #139, opened 2019-06-19
ipmitool's dependency on gettext


siddharthwagh:
ipmitool has a dependency on gettext and related packages due to use of eval_gettext following two scripts:
- contrib/exchange-bmc-os-info.init.redhat
- contrib/bmc-snmp-proxy

The former script attempts to workaround use of eval_gettext by adding disabling GETTEXT flag if gettext is not installed. But the latter does not have the same approach. Additionally, the location of gettext.sh file is incorrect.

PROPOSED FIXES:

- Use the correct location of gettext.sh file as /usr/bin/gettext.sh
- Apply the eval_gettext workaround in both scripts to make the requirement of gettext package optional for ipmitool
`
if [ -f /usr/bin/gettext.sh ]; then
   GETTEXT=1
   . /usr/bin/gettext.sh
   OUTPUT="eval_gettext"
else
   GETTEXT=0
   OUTPUT="echo"
fi 
`

PR #140, opened 2019-06-19
Remove gettext dependency and correct script path

siddharthwagh:
These changes make ipmitool's dependency requirement of gettext and related packages like cvs optional. They also correct the path of gettext.sh script from /bin/gettext.sh to /usr/bin/gettext.sh. As a result, ipmitool package would not need to include gettext as a requirement under spec file, but it will be used if already installed on a system.

Resolves ipmitool/ipmitool#139
Signed-off-by: Siddharth Wagh <swagh@proofpoint.com>

### DIFF
--- a/contrib/bmc-snmp-proxy
+++ b/contrib/bmc-snmp-proxy
@@ -44,7 +44,14 @@ RELOAD_SERVICES="yes"
 # source config
 [ -r ${CONFIG} ] && . ${CONFIG}
 
-. gettext.sh
+if [ -f /usr/bin/gettext.sh ]; then
+	GETTEXT=1
+	. /usr/bin/gettext.sh
+	OUTPUT="eval_gettext"
+else
+	GETTEXT=0
+	OUTPUT="echo"
+fi
 
 SCRIPT_NAME=$(basename $0)
 RETVAL=0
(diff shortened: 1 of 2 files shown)


contrib/exchange-bmc-os-info.init.redhat
```diff
@@ -64,9 +64,9 @@ SCRIPT_NAME=$(basename $0)
 
 RETVAL=0
 
-if [ -f /bin/gettext.sh ]; then
+if [ -f /usr/bin/gettext.sh ]; then
```
AlexanderAmelkin: Although this is redhat-specific, why not search in PATH like I proposed for the other script?

contrib/bmc-snmp-proxy
```diff
@@ -379,22 +387,23 @@ esac
...
-	5) eval_gettext "${SCRIPT_NAME}: failed to disable snmp proxy." 1>&2 ;;
-	6) eval_gettext "${SCRIPT_NAME}: failed to reload snmpd." 1>&2 ;;
-	7) eval_gettext "${SCRIPT_NAME}: failed to set snmpd config." 1>&2 ;;
-	8) eval_gettext "${SCRIPT_NAME}: failed to set IPMI alert dest." 1>&2 ;;
-	9) eval_gettext "${SCRIPT_NAME}: no free IPMI alert dest." 1>&2 ;;
-	10) eval_gettext "${SCRIPT_NAME}: failed to set IPMI PEF." 1>&2 ;;
-	11) eval_gettext "${SCRIPT_NAME}: failed to write snmptrapd.conf." 1>&2 ;;
-	12) eval_gettext "${SCRIPT_NAME}: snmpd not found." 1>&2 ;;
-	*) eval_gettext "${SCRIPT_NAME}: unknown error." 1>&2 ;;
+	2) ${OUTPUT} "${SCRIPT_NAME}: failed to read ${BMC_INFO} " 1>&2 ;;
```
AlexanderAmelkin: I believe `gettext` message ids must not include variable parts. Also, I don't see `TEXTDOMAIN` being set anywhere, so `gettext` won't probably find anything. Especially provided that there are no i18n files. Does anyone actually use these scripts? 

I know it wasn't your code initially, but since you've decided to make these scripts work, please make them actually work. That being said, I think that neither `"${SCRIPT_NAME}: "`, nor `"${BMC_INFO"` or whatever other variables must be left in the messages. For compatibility with various languages, I would change the output routine to `printf` or a `gettext`-obtained value with format specifiers. Like this:

```
output() {
    FMT=$(${OUTPUT} $1)
    shift
    printf "${FMT}" $*
}

...

output "Some fancy message about entity's %s value %d\n" ${ENTITY} ${VALUE}
```

AlexanderAmelkin:
@siddharthwagh, hello?

AlexanderAmelkin:
@siddharthwagh, are you still willing to fix this?

PR #142, opened 2019-06-25
fix embedded bridged responses handling

### LINKED ISSUE #141: incorrect handling of embedded bridged responses
when receiving an embedded bridged response, the following fields are not being updated:
- offset,
- payload_start, 
- payload_size,

which results in parsing improper data as the header of the embedded response.

dfrycki:
fixes #141

### DIFF
--- a/src/plugins/lanplus/lanplus.c
+++ b/src/plugins/lanplus/lanplus.c
@@ -762,12 +762,6 @@ ipmi_lan_poll_single(struct ipmi_intf * intf)
 						printbuf(&rsp->data[offset], (rsp->data_len-offset-1),
 								"bridge command response");
 						/*
-						 * decrement payload size
-						 * (cks2 for outer Send Message)
-						 */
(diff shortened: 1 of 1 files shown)


vmauery:
I helped @dfrycki debug this issue on an OpenBMC platform. When the memmove happens, it was not correctly resetting the offset and shadow offset into that buffer, causing out-of-bounds memory references for encapsulated responses.

AlexanderAmelkin:
@dfrycki, @vmauery, have you checked it on non-OpenBMC stacks? I don't have any systems with bridging to verify it myself.

vmauery:
I have tested it on two generations of prior Intel systems, both of which are different IPMI stacks (AMI-based and proprietary). From what we can understand from the behavior of those is that they split the response into two packets and skip this broken path. OpenBMC sends the response all in one packet and hits this issue where the offsets are incorrect.

But now either path works.

PR #153, opened 2019-09-02
Improve clarity of the "set password" length argument

gdttn:
Some small changes to the tool documentation / man page to make it clearer how the password length argument is handled.

### DIFF
--- a/lib/ipmi_user.c
+++ b/lib/ipmi_user.c
@@ -657,24 +657,25 @@ ipmi_user_password(struct ipmi_intf *intf, int argc, char **argv)
 		}
 	} else {
 		password = argv[3];
-		if (argc > 4) {
-			if ((str2uchar(argv[4], &password_type) != 0)
-					|| (password_type != 16 && password_type != 20)) {
-				lprintf(LOG_ERR, "Invalid password length '%s'", argv[4]);
-				return (-1);
-			}
-		} else {
-			password_type = 16;
-		}
 	}
+	
+        if (argc > 4) {
+                if ((str2uchar(argv[4], &password_type) != 0)
+                                || (password_type != 16 && password_type != 20)) {
+                        lprintf(LOG_ERR, "Invalid password length '%s'", argv[4]);
+                        return (-1);
+                }
+        } else if (strlen(password) > 16) {
+                password_type = 20;
+        }
 
-	if (!password) {
-		lprintf(LOG_ERR, "Unable to parse password argument.");
-		return (-1);
-	} else if (strlen(password) > 20) {
-		lprintf(LOG_ERR, "Password is too long (> 20 bytes)");
-		return (-1);
-	}
+        if (!password) {
+                lprintf(LOG_ERR, "Unable to parse password argument.");
+                return (-1);
+        } else if (strlen(password) > password_type) {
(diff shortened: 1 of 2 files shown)


lib/ipmi_user.c
```diff
@@ -434,7 +434,7 @@ print_user_usage(void)
 	lprintf(LOG_NOTICE,
 "               set name     <user id> <username>");
 	lprintf(LOG_NOTICE,
-"               set password <user id> [<password> <16|20>]");
+"               set password <user id> [<password> [<16|20>]]");
```
AlexanderAmelkin: Don't you want to also change the code so that if the actual password string length is longer than 16, it would automatically set length to 20 ? So far, as I can see, it assumes 16 for both command line and stdin passwords, but always uses 20 when checking for the length limit. Doesn't look correct to me. If you're not willing to fix that, it's ok, I can do it myself later.

gdttn: Thanks for looking it over. I agree the length handling is a bit odd but I wasn't sure that detecting the size was necessarily "correct" or consistent behaviour.  I'd be happy to implement it at your suggestion.

At the moment it will return an error when the password is over 20 characters, but that is at odds with the truncation behaviour for a password of length 16.

I'd suggest that the code should behave consistently: if the password length exceeds an _explicit_ length, that seems like an error condition.  But if no length is specified, it would be OK to automatically detect (and raise an error if the max length is exceeded).

(But isn't the whole thing complicated by the fact that 20-character passwords aren't accepted over some communications channels?  Or did I misremember that?)

gdttn:
Done, hope that helps.

There are probably still some inconsistencies in documentation and code WRT IPMI1.x/2.0 handling. The man page says,

> For IPMI v1.5, the maximum password length is 16 characters.
> Passwords longer than 16 characters will be truncated.
> 
> For IPMI v2.0, the maximum password length is 20 characters;
> longer passwords are truncated.

And given that the tool does _some_ length checks at the user side (just not protocol-specific ones) this statement isn't entirely true of `ipmitool` either.

lib/ipmi_user.c
AlexanderAmelkin: I suppose the BMC will return an error if a >16 character password is attempted when IPMI 2.0 isn't supported. Hence, I think the behavior you propose would be just fine:

> I'd suggest that the code should behave consistently: if the password length exceeds an explicit length, that seems like an error condition. But if no length is specified, it would be OK to automatically detect (and raise an error if the max length is exceeded).

AlexanderAmelkin:
I've fixed some formatting and also dropped an extra check for strlen when reading the passwords from stdin as strncmp already does length compare. Please check that it works for you. I also think that the place in the doc that you've mentioned (regarding truncation) must be updated as well, but I haven't done that.

gdttn:
Everything works for me, thanks!

I've update the manual page again as you suggest.  I don't see that ipmitool does any checking of whether the interface is compatible with the password length, it just throws it at the remote system and waits for a result.  I also don't have a 1.5/lan interface to test with, so can't do much more to check.

I've done all I can to reflect this uncertainty, and corrected the description of the 2.0 behaviour.

gdttn:
Excellent, thank you for your help!

PR #157, opened 2019-09-20
Add support for specifying network source address

oko:
This PR adds the ability to specify a source address to use for hosts which have multiple useable addresses.


AlexanderAmelkin:
Isn't this supposed to be governed by system routing rules? Like
`ip ro add 10.0.0.2 dev br0 src 10.0.0.1` ?

Issue #163, opened 2019-11-04
Unable to set options=efiboot,persistent


jvillal-amp:
I'm thinking the parsing of multiple "options=" values is somehow broken.

When I do the following:
```
ipmitool -I lanplus chassis bootdev disk options=persistent,efiboot -vvv
```

I see the following
```
<snip>
Chassis Set Boot Parameter 4 to 0101

>> Sending IPMI command payload
>>    netfn   : 0x00
>>    command : 0x08
>>    data    : 0x05 0xa0 0x08 0x00 0x00 0x00
<snip>
```
Notice the data is ``0x05 0xa0 ...``  0xA0 means EFI but NOT persistent.

When doing the command:
```
ipmitool -I lanplus chassis bootdev disk options=efiboot,persistent -vvv
```

I see:
```
<snip>
Chassis Set Boot Parameter 4 to 0101

>> Sending IPMI command payload
>>    netfn   : 0x00
>>    command : 0x08
>>    data    : 0x05 0xc0 0x08 0x00 0x00 0x00
<snip>
```
Notice the data is ``0x05 0xc0 ...``   0xC0 means persistent but NOT EFI

What we desire should look like this for either call:
```
Chassis Set Boot Parameter 4 to 0101

>> Sending IPMI command payload
>>    netfn   : 0x00
>>    command : 0x08
>>    data    : 0x05 0xe0 0x08 0x00 0x00 0x00
```
The data should be ``0x05 0xe0 ...``  0xE0 means EFI and persistent

The values are specified in Table 28-14 `Boot Option Parameters` in the IPMI Specification, V2.0, Rev. 1.1 https://www.intel.com/content/dam/www/public/us/en/documents/product-briefs/ipmi-second-gen-interface-spec-v2-rev1-1.pdf

AlexanderAmelkin:
Whoa! There is a whole nest of bugs in that parser and its counterpart!

The main bug that causes this particular issue is in line 1773 of ipmi_chassis.c where it says `flags[op->i] &= op->mask` instead of `flags[op->i] &= ~(op->mask)`, thus wiping out all other flags instead of clearing the state of the flag being set.

In addition to that, the code also confuses console redirection flags with boot verbosity flags and chokes on remote/redirected boot media flags.

I'm going to fix all that.

qthangnp:
When will the new ipmitool 1.8.19 released?

djuarezg:
I guess this has not yet been released, hasn't it?

[...]

PR #164, opened 2019-11-04
Fix options= parsing for persistent and efiboot

jvillal-amp:
When using 'persistent' and/or 'efiboot' in an 'options=' like this:
  ipmitool -I lanplus chassis bootdev disk options=efiboot,persistent

Allow them to be used together. Previously if you specified them
together only the last one specified would take effect. That would
mean you could either specify to set for EFI or you could set it to be
persistent. But could not set EFI and persistent.

Changed the mask values so that they no longer wipe out previous
values.

Resolves: ipmitool#163

### DIFF
--- a/lib/ipmi_chassis.c
+++ b/lib/ipmi_chassis.c
@@ -1714,9 +1714,9 @@ ipmi_chassis_main(struct ipmi_intf * intf, int argc, char ** argv)
 					/* data 1 */
 					{"valid", 0, (1<<7), (1<<7),
 						"Boot flags valid"},
-					{"persistent", 0, (1<<6), (1<<6),
+					{"persistent", 0, (0xe0), (1<<6),
 						"Changes are persistent for all future boots"},
-					{"efiboot", 0, (1<<5), (1<<5),
+					{"efiboot", 0, (0xe0), (1<<5),
 						"Extensible Firmware Interface Boot (EFI)"},
 					/* data 2 */
 					{"clear-cmos", 1, (1<<7), (1<<7),


lib/ipmi_chassis.c
```diff
@@ -1714,9 +1714,9 @@ ipmi_chassis_main(struct ipmi_intf * intf, int argc, char ** argv)
 					/* data 1 */
 					{"valid", 0, (1<<7), (1<<7),
 						"Boot flags valid"},
-					{"persistent", 0, (1<<6), (1<<6),
+					{"persistent", 0, (0xe0), (1<<6),
 						"Changes are persistent for all future boots"},
-					{"efiboot", 0, (1<<5), (1<<5),
+					{"efiboot", 0, (0xe0), (1<<5),
```
AlexanderAmelkin: Sorry, this fix is plain wrong. The bug is not there. It is in line 1773 where it says `flags[op->i] &= op->mask` instead of `flags[op->i] &= ~(op->mask)`, thus wiping out all other flags instead of clearing the state of the flag being set. The table must contain 1s in mask for the bits that need to be cleared when a particular option is being set. Also, this code (the original one) needs to be cleaned of magic numbers.

AlexanderAmelkin:
Sorry, I won't merge this. I will fix #163 myself though in a different PR.

lib/ipmi_chassis.c
jvillal-amp: Yeah, I wasn't sure about that. I thought about the ``flags[op->i] &= ~(op->mask)`` but then I was scared I would break something else. So I went for the simplest "fix" :) Thanks!

jvillal-amp:
> Sorry, I won't merge this. I will fix #163 myself though in a different PR.

Spasibo Alexander! Next time I am in Moscow I should buy you a coffee/drink :)

AlexanderAmelkin:
You're welcome. Please check #165.

PR #165, opened 2019-11-06
Fix/cleanup chassis bootparam and bootdev options

### LINKED ISSUE #163: Unable to set options=efiboot,persistent

AlexanderAmelkin:
This resolves #163 and also cleans up `chassis bootparam get` decoding and `chassis bootdev options=...` parsing.

Please read individual commit messages.

### DIFF
--- a/lib/ipmi_chassis.c
+++ b/lib/ipmi_chassis.c
@@ -1712,6 +1712,205 @@ ipmi_chassis_set_bootflag_help()
 	get_bootparam_options("options=help", &set_flag, &clr_flag);
 }
 
+/*
+ * Sugar. Macros for internal use by bootdev_parse_options() to make
+ * the structure initialization look better
+ */
+#define BF1_OFFSET 0
+#define BF2_OFFSET 1
+#define BF3_OFFSET 2
+#define BF4_OFFSET 3
+#define BF_BYTE_COUNT 5
+
+/* A helper for ipmi_chassis_main() to parse bootdev options */
+static
+bool
+bootdev_parse_options(char *optstring, uint8_t flags[])
+{
+	char *token;
+	char *saveptr = NULL;
+	int optionError = 0;
+	static struct {
+		char *name;
+		off_t offset;
+		unsigned char mask;
+		unsigned char value;
+		char *desc;
+	} *op, options[] = {
+		/* data 1 */
+		{
+			"valid",
+			BF1_OFFSET,
+			BF1_VALID_MASK,
+			BF1_VALID,
+			"Boot flags valid"
+		},
+		{
+			"persistent",
+			BF1_OFFSET,
+			BF1_PERSIST_MASK,
+			BF1_PERSIST,
+			"Changes are persistent for "
+				"all future boots"
+		},
+		{
+			"efiboot",
+			BF1_OFFSET,
+			BF1_BOOT_TYPE_MASK,
+			BF1_BOOT_TYPE_EFI,
+			"Extensible Firmware Interface "
+				"Boot (EFI)"
+		},
+		/* data 2 */
+		{
+			"clear-cmos",
+			BF2_OFFSET,
+			BF2_CMOS_CLEAR_MASK,
+			BF2_CMOS_CLEAR,
+			"CMOS clear"
+		},
+		{
+			"lockkbd",
+			BF2_OFFSET,
+			BF2_KEYLOCK_MASK,
+			BF2_KEYLOCK,
+			"Lock Keyboard"
+		},
+		/* data2[5:2] is parsed elsewhere */
+		{
+			"screenblank",
+			BF2_OFFSET,
+			BF2_BLANK_SCREEN_MASK,
+			BF2_BLANK_SCREEN,
+			"Screen Blank"
+		},
+		{
+			"lockoutreset",
+			BF2_OFFSET,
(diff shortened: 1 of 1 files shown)


AlexanderAmelkin:
@jvillal-amp, please check if this works for you.

lib/ipmi_chassis.c
```diff
@@ -1712,6 +1712,205 @@ ipmi_chassis_set_bootflag_help()
...
 
+/*
+ * Sugar. Macros for internal use by bootdev_parse_options() to make
+ * the structure initialization look better
+ */
+#define BF1_OFFSET 0
+#define BF2_OFFSET 1
+#define BF3_OFFSET 2
+#define BF4_OFFSET 3
+#define BF_BYTE_COUNT 5
```
jvillal-amp: Random thought: If they were like 
```
static const int BF1_OFFSET = 0
```

Then they would be contained within the scope of this function and also type safe.

But also good to leave as is.

lib/ipmi_chassis.c
```diff
@@ -1712,6 +1712,205 @@ ipmi_chassis_set_bootflag_help()
...
+
+/* A helper for ipmi_chassis_main() to parse bootdev options */
+static
+bool
+bootdev_parse_options(char *optstring, uint8_t flags[])
+{
+	char *token;
+	char *saveptr = NULL;
+	int optionError = 0;
+	static struct {
```
vmauery: static const struct?

lib/ipmi_chassis.c
```diff
@@ -1700,98 +2057,13 @@ ipmi_chassis_main(struct ipmi_intf * intf, int argc, char ** argv)
...
-					}
-				}
-				if (!op->name) {
-					/* Option not found */
-					optionError = 1;
-					lprintf(LOG_ERR, "Invalid option: %s", token);
-				}
-				token = strtok_r(NULL, ",", &saveptr);
+				uint8_t flags[BF_BYTE_COUNT];
+				if (!bootdev_parse_options(argv[2] + 8, flags))
```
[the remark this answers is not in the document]
JohnVillalovos: > 
> 
> What is 8?

The length of "options=", but agree that it isn't obvious.

lib/ipmi_chassis.c
AlexanderAmelkin: Good point

lib/ipmi_chassis.c
```diff
@@ -1554,6 +1712,205 @@ ipmi_chassis_set_bootflag_help()
...
+	char *token;
+	char *saveptr = NULL;
+	int optionError = 0;
+	static struct {
+		char *name;
+		off_t offset;
+		unsigned char mask;
+		unsigned char value;
+		char *desc;
+	} *op, options[] = {
```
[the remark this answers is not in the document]
AlexanderAmelkin: Not my code (I only moved `*op` before `options`), but I agree.

lib/ipmi_chassis.c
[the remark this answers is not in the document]
AlexanderAmelkin: I agree. Hate magic numbers myself. Not my `8` originally, but I though in this case where it is adjacent to `strcmp` and the `"options="` string, it's not that of a crime. I guess a comment would suffice. Will add one.

lib/ipmi_chassis.c
AlexanderAmelkin: Agreed

lib/ipmi_chassis.c
AlexanderAmelkin: Alas, it turned out that gcc5 (default for Ubuntu Xenial) can't use `static const int` for structure initialization. gcc8 can do that, but for compatibility I'll keep these as macros.

PR #168, opened 2019-11-12
Docs: Add info on packages to install on Ubuntu 16.04 and Fedora 31

jvillal-amp:
Add some documenation on packages to install on Ubuntu 16.04 before
building the software.

This will make it easier for people who are building ipmitool as they
won't need to figure out all the package dependencies on their own.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -22,6 +22,12 @@ project is tested automatically to build cleanly for the following
    - Ubuntu 16.04 Xenial
    - MacOS X 10.14 (Xcode 10.2)
 
+   For Ubuntu 16.04 it is recommended to do the following before building:
+$ sudo apt install automake gcc git libncurses-dev libreadline-dev libssl-dev libsystemd-dev libtool make wget
+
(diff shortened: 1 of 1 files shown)


INSTALL
```diff
@@ -29,6 +29,9 @@ Basic Installation
 install this package:
    `./bootstrap && ./configure && make && make install`
 
+   For Ubuntu 16.04 it is recommended to do the following before building:
+$ sudo apt install git automake curl gcc libncurses-dev libreadline-dev libssl-dev libsystemd-dev libtool make wget
```
AlexanderAmelkin: It's actually either `curl` or `wget`, not both.
Also, it would be better to put the prerequisite info above the installation instruction.

jvillal-amp: > It's actually either `curl` or `wget`, not both.

Okay. I thought it didn't hurt to have both 😊

> Also, it would be better to put the prerequisite info above the installation instruction.

Will do!

Thanks.

PR #172, opened 2019-11-18
Add support for editing FRU 'board mfg date'.

slashwu:
This PR adds the ability to edit FRU 'board mfg date'.
Tested as screenshot.
edit_fru_board_mfg_date

### DIFF
--- a/lib/ipmi_fru.c
+++ b/lib/ipmi_fru.c
@@ -4767,6 +4767,45 @@ f_type, uint8_t f_index, char *f_string)
 	}
 	/* Convert index from character to decimal */
 	f_index= f_index - 0x30;
+	
+	if (f_type == 'b' && f_index == 9)		// use index '9' for editing 'board mfg date', by slash.wu@ztsystems.com
+	{
+		int y, m, d, h, min;
+		struct tm board_mfg_date;
+		time_t total_secs;
+	
+		sscanf(f_string, "%4d%2d%2d%2d%2d", &y, &m, &d, &h, &min);
+		board_mfg_date.tm_year = y - 1900;
+		board_mfg_date.tm_mon = m - 1;
+		board_mfg_date.tm_mday = d;
+		board_mfg_date.tm_hour = h;
+		board_mfg_date.tm_min = min;
+		board_mfg_date.tm_sec = 0;
+		total_secs = mktime(&board_mfg_date);
+		total_secs -= secs_from_1970_1996;
+		total_secs /= 60;			// mins
+	
+		memcpy(&fru_data[3], &total_secs, 3);
+	
+		checksum = 0;
+	/* Calculate Header Checksum */
+		for (i = 0; i < fru_section_len - 1; i++)
+		{
+			checksum += fru_data[i];
+		}
+		checksum = (~checksum) + 1;
+		fru_data[fru_section_len - 1] = checksum;
+	
+	/* Write the updated section to the FRU data; source offset => 0 */
+		if (write_fru_area(intf, &fru, fruId, 0, header_offset, fru_section_len, fru_data) < 0)
+		{
+			printf("Write to FRU data failed.\n");
+			rc = (-1);
+			goto ipmi_fru_set_field_string_out;
+		}
+		else
+			goto ipmi_fru_set_field_string_out;
+	}
 
 	/*Seek to field index */
 	for (i=0; i <= f_index; i++) {


lib/ipmi_fru.c
```diff
@@ -4767,6 +4767,46 @@ f_type, uint8_t f_index, char *f_string)
 	}
 	/* Convert index from character to decimal */
 	f_index= f_index - 0x30;
+	
+	if (f_type == 'b' && f_index == 9)		// use index '9' for editing 'board mfg date', by slash.wu@ztsystems.com
+	{
+		const uint64_t secs_from_1970_1996 = 820454400;
```
AlexanderAmelkin: This is already defined in `ipmi_fru.h` in `ipmi_fru2time_t()`.
Make it a macro within that header and move the appropriate code there into a new `ipmi_time2fru(time_t *, void *mfg_date)`.

lib/ipmi_fru.c
```diff
@@ -4767,6 +4767,46 @@ f_type, uint8_t f_index, char *f_string)
...
+		board_mfg_date.tm_mon = m - 1;
+		board_mfg_date.tm_mday = d;
+		board_mfg_date.tm_hour = h;
+		board_mfg_date.tm_min = min;
+		board_mfg_date.tm_sec = 0;
+		total_secs = mktime(&board_mfg_date);
+		total_secs -= secs_from_1970_1996;
+		total_secs /= 60;			// mins
+	
+		memcpy(&fru_data[3], &total_secs, 3);
```
AlexanderAmelkin: Nope. This will only work on x86 and other little-endian systems. You must use `htoipmi24()` instead of `memcpy()`.

lib/ipmi_fru.c
```diff
@@ -4767,6 +4767,46 @@ f_type, uint8_t f_index, char *f_string)
...
+		fru_data[fru_section_len - 1] = checksum;
+	
+	/* Write the updated section to the FRU data; source offset => 0 */
+		if (write_fru_area(intf, &fru, fruId, 0, header_offset, fru_section_len, fru_data) < 0)
+		{
+			printf("Write to FRU data failed.\n");
+			rc = (-1);
+			goto ipmi_fru_set_field_string_out;
+		}
+		else
```
AlexanderAmelkin: No need to repeat the same `goto` twice. No need for the `else`.

lib/ipmi_fru.c
```diff
@@ -4767,6 +4767,46 @@ f_type, uint8_t f_index, char *f_string)
...
+		sscanf(f_string, "%4d%2d%2d%2d%2d", &y, &m, &d, &h, &min);
+		board_mfg_date.tm_year = y - 1900;
+		board_mfg_date.tm_mon = m - 1;
+		board_mfg_date.tm_mday = d;
+		board_mfg_date.tm_hour = h;
+		board_mfg_date.tm_min = min;
+		board_mfg_date.tm_sec = 0;
+		total_secs = mktime(&board_mfg_date);
+		total_secs -= secs_from_1970_1996;
+		total_secs /= 60;			// mins
```
AlexanderAmelkin: Please create a counterpart for the existing `ipmi_strftime()` (see `ipmi_time.h`). The proposed name is `ipmi_strptime()`. Don't reinvent the bicycle with an `sscanf()` and manual field assignments, use `strptime()` and `mktime()` to get a `time_t` (seconds) value from a string.

AlexanderAmelkin:
I see that you've created a branch that addresses some of the review comments. Please instead, to preserve the review history, force-push to your master as it is linked to this PR.

lib/ipmi_fru.c
```diff
@@ -4564,7 +4564,7 @@ ipmi_fru_main(struct ipmi_intf * intf, int argc, char ** argv)
 					ipmi_fru_edit_help();
 					return -1;
 				}
-				rc = ipmi_fru_set_field_string(intf, fru_id, *argv[3], *argv[4],
+				rc = ipmi_fru_set_field_string(intf, fru_id, *argv[3], atoi(argv[4]),  // the 4th param was changed to convert w/ atoi() by slash
```
AlexanderAmelkin, line 4567: `atoi()` returns 0 on conversion errors. can't use it here. generally, try to avoid using `atoi()`, use `strtol()` instead.

AlexanderAmelkin: no need to put your name in a comment near each change. there is `git blame` for that.

lib/ipmi_fru.c
```diff
@@ -4766,7 +4766,7 @@ f_type, uint8_t f_index, char *f_string)
 		goto ipmi_fru_set_field_string_out;
 	}
 	/* Convert index from character to decimal */
-	f_index= f_index - 0x30;
+	//f_index= f_index - 0x30;  // comment out since the caller changed to atoi(), by slash
```
AlexanderAmelkin, line 4769: no need to comment out unneeded lines, just delete them

lib/ipmi_fru.c
```diff
@@ -5074,7 +5074,7 @@ ipmi_fru_set_field_string_rebuild(struct ipmi_intf * intf, uint8_t fruId,
 			header.offset.board   += change_size_by_8;
 		}
 		/* Board type field */
-		if ((f_type == 'c' ) || (f_type == 'b' ))
+		if (((f_type == 'c' ) || (f_type == 'b' )) && header.offset.product)
```
AlexanderAmelkin, line 5077: this line looked wrong before your change, and with your change it looks even more wrong.
the code below is about product section, why at all check for 'b' (board) and 'c' (chassis) type while it must be 'p' (product) instead?

lib/ipmi_main.c
```diff
@@ -845,7 +845,7 @@ ipmi_main(int argc, char ** argv,
 	log_init(progname, 0, verbose);
 
 	/* load the IANA PEN registry */
-	if (ipmi_oem_info_init()) {
+	if (0 && ipmi_oem_info_init()) {  // not do init by slash
```
AlexanderAmelkin, line 848: this is definitely NOT going to be merged. please avoid debug changed being added to a PR.

AlexanderAmelkin (changes requested):

AlexanderAmelkin:
@doublechiang, don't you want to make a PR out of your branch? The author of this PR doesn't look interested in fixing it and just uses the branch for some experiments as it looks.

PR #178, opened 2019-12-04
WIP: Add gihub pages site contents

AlexanderAmelkin:
We need a place to put announcements and things like that. Github Pages looks suitable for that. This PR is to add the site contents usable by github.io.


AlexanderAmelkin:
> Another alternative would be to just move the manpages into the main dir, but this works too.

Yep. Actually the directory structure of the project is quite strange. For instance, there is this `lib` part that doesn't actually constitute a usable library and so it's hard to understand why `lib` and `src` are separated. The `src/plugins` name is also misleading because they aren't actually plugins, but are in fact compile-in modules.

Nonetheless, I think that putting manpages into `man` is ok.

AlexanderAmelkin:
This is not needed anymore as GitHub has changed how Pages work.

PR #183, opened 2020-02-17
ipmi_sel_set_time: fix strptime() return check

PavelKiryukhin:
The current behavior:
- correct date format is not accepted by "sel time set".
- incorrect date format that looks correct is accepted, but time is not
  set correctly.
- commands like
    ipmitool sel time set "11/22/2013 trash"
  are accepted.


PavelKiryukhin:
Hi, is it possible to review this patch?
My company (Kontron) want it to be merged as upstream sel set time does not work w/o this commit.
--
Thank you,
Pavel Kiryukhin

AlexanderAmelkin:
Sorry for the delay.

PR #186, opened 2020-02-23
Use -n to omit timestamps in manpages for reproducablity

jpuhlman:


Makefile.am
```diff
@@ -85,7 +85,7 @@ pkg: control/pkginfo
 	$(STRIP) src/ipmitool
 	(cd control ; pkgmk -o -d /tmp)
 	pkgtrans -s /tmp $(PKG) ipmitool
-	gzip -c /tmp/$(PKG) > $(PKG).gz
+	gzip -nc /tmp/$(PKG) > $(PKG).gz
```
AlexanderAmelkin: Are you really using Solaris? As far as I can see, the whole `pkg` target is non-functional on Linux (at least Ubuntu) because there are no `pkgmk` and `pkgtrans` tools.

Is this really an issue on Solaris? Is this the only thing that prevents reproducible builds for you?

jpuhlman: No, I am working on yocto based linux in this case.

AlexanderAmelkin: Well, not that I'm totally against this change, but I'm still not convinced that it is actually needed. How come that a Yocto-based Linux uses this `Makefile` target? In the default openembedded recipe for `ipmitool` I do not see any references to this target, as well as I don't see any recipe that provides `pkgtrans` that is required for this target.

It still looks like this target is some kind of atavism and needs to be removed rather than patched.

Am I wrong?

PR #191, opened 2020-04-02
read_fru_area: fix memset() parameters

PavelKiryukhin:
Current memset parameters are calculated in the assumption
that we use the buffer for whole FRU. Whereas we use individual
buffers for each area that are smaller than whole FRU size.
That may lead to attempt to memset beyond allocated buffer.
Also: add comment to fail path in fru_area_print_product().

### DIFF
--- a/lib/ipmi_fru.c
+++ b/lib/ipmi_fru.c
@@ -679,11 +679,11 @@ read_fru_area(struct ipmi_intf * intf, struct fru_info *fru, uint8_t id,
 
 	finish = offset + length;
 	if (finish > fru->size) {
-		memset(frubuf + fru->size, 0, length - fru->size);
-		finish = fru->size;
+		memset(frubuf + fru->size - offset, 0, finish - fru->size);
 		lprintf(LOG_NOTICE, "Read FRU Area length %d too large, "
 			"Adjusting to %d",
-			offset + length, finish - offset);
+			finish - offset, fru->size - offset);
+		finish = fru->size;
 		length = finish - offset;
 	}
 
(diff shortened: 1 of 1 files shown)


yontalcar:
I think it would be more straightforward to zero the whole buffer unconditionally as there are more cases where the buffer won't be filled (https://github.com/ipmitool/ipmitool/blob/master/lib/ipmi_fru.c#L777). Or drop the memset call there as the buffer should be zeroed by calling function.

Also the same buggy code is in read_fru_area_section().

AlexanderAmelkin:
The whole fru support is rusty, buggy, spaghetti-like and generally awful. It needs a complete re-write or at least a major refactoring. I've been thinking about it for a long time already. Thought of maybe expanding the [frugen](https://github.com/ipmitool/frugen) libfru library and using it, thought of other approaches, but I still don't have time and courage to start this fight.

These minor fixes may be fine, but they are doing something with all these strange functions that duplicate one another, and use strange names for their arguments and variables, so I am not sure how to review this and be sure that the fixes are correct. When I start doing so, I immediately rediscover how awful the fru code is and feel the urge to rewrite it all, but... (see above).

Hopefully one day soon I will find time to rewrite it all, or maybe some Heracles appears to help me clean these Augean stables of fru... Until then I think I will not be accepting fixes to fru. Sorry.

yontalcar:
@AlexanderAmelkin
I agree that the code is ugly and confusing, but this is a regression from [IPMITOOL_1_8_18](https://github.com/ipmitool/ipmitool/releases/tag/IPMITOOL_1_8_18) introduced in commit e824c23316ae50beb7f7488f2055ac65e8b341f2 .

So I guess something should be done about it.

pcahyna:
@PavelKiryukhin is there a reproducer to show some problematic behavior? While the code does not look correct, `ipmitool fru print` seems to behave as it should.

PR #197, opened 2020-04-16
harden ipmi_sdr_get_record() against bad records

jlevon:
On one of our production systems (Supermicro X9DRD-7LN4F based), ipmitool sensor was occasionally crashing. This was due to it sometimes returning a junk SDR of the form:

```
SDR record ID   : 0x008a
SDR record id mismatch: 0x000a
SDR record type : 0x00
SDR record next : 0x0000
SDR record bytes: 100
```

or something similar. In this case, when we request the SDR data, we're getting back 0xca. However, ipmi_sdr_get_record() can underflow sdr_max_read_len to a negative value in this case. Harden the code for this.

Also, fix the addition of i to use the correct amount: it doesn't actually matter, since we'll check for i < len in the loop condition, but still.

### DIFF
--- a/lib/ipmi_sdr.c
+++ b/lib/ipmi_sdr.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2012 Hewlett-Packard Development Company, L.P.
+ * Copyright 2020 Joyent, Inc.
  *
  * Based on code from
  * Copyright (c) 2003 Sun Microsystems, Inc.  All Rights Reserved.
@@ -3159,7 +3160,8 @@ ipmi_sdr_get_record(struct ipmi_intf * intf, struct sdr_get_rs * header,
 			sdr_rq.length, sdr_rq.offset);
 
 		rsp = intf->sendrecv(intf, &req);
-		if (!rsp) {
+
+		if (!rsp || rsp->ccode == 0xca) {
 		    sdr_max_read_len = sdr_rq.length - 1;
 		    if (sdr_max_read_len > 0) {
 			/* no response may happen if requests are bridged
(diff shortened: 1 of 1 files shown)


lib/ipmi_sdr.c
```diff
@@ -3159,7 +3160,8 @@ ipmi_sdr_get_record(struct ipmi_intf * intf, struct sdr_get_rs * header,
 			sdr_rq.length, sdr_rq.offset);
 
 		rsp = intf->sendrecv(intf, &req);
-		if (!rsp) {
+
+		if (!rsp || rsp->ccode == 0xca) {
```
AlexanderAmelkin, line 3164: This is `IPMI_CC_CANT_RET_NUM_REQ_BYTES` from `ipmitool/ipmi_cc.h`. Please don't use magic numbers.

lib/ipmi_sdr.c
```diff
@@ -3170,14 +3172,7 @@ ipmi_sdr_get_record(struct ipmi_intf * intf, struct sdr_get_rs * header,
...
 		    }
-		}
-
-		switch (rsp->ccode) {
-		case 0xca:
-			/* read too many bytes at once */
-			sdr_max_read_len = sdr_rq.length - 1;
-			continue;
-		case 0xc5:
+		} else if (rsp->ccode == 0xc5) {
```
AlexanderAmelkin, line 3175: `IPMI_CC_RES_CANCELED`

lib/ipmi_sdr.c
jlevon: I couldn't agree more (and I fixed it up), but to be fair, this is what the existing code did.

jlevon:
Thanks for reviewing @AlexanderAmelkin

PR #198, opened 2020-04-16
define non-portable MAX() if not defined in ipmi_hpmfwupg.c

jlevon:
Relying on sys/param.h's MAX() definition is not portable.

### DIFF
--- a/lib/ipmi_hpmfwupg.c
+++ b/lib/ipmi_hpmfwupg.c
@@ -49,10 +49,6 @@
 # include <config.h>
 #endif
 
-#ifndef MAX
-#define MAX(a,b) ((a) > (b) ? (a) : (b))
-#endif
-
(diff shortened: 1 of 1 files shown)


lib/ipmi_hpmfwupg.c
```diff
@@ -48,6 +49,10 @@
 # include <config.h>
 #endif
 
+#ifndef MAX
+#define MAX(a,b) ((a) > (b) ? (a) : (b))
```
AlexanderAmelkin, line 53: There is already `__max()` and `__min()` macros in  `#include <ipmitool/helper.h>`. Please change to using that instead of defining `MAX` here.

jlevon: Thanks, updated.

PR #202, opened 2020-05-23
fixes #104, Clean up use of strcmp/strncmp/strncasecmp for command li…

### LINKED ISSUE #104: Clean up use of strcmp/strncmp/strncasecmp for command line arguments
This is associated with [SourceForge ticket 24](https://sourceforge.net/p/ipmitool/bugs/24/).
In many places command line arguments are matched using some weird constructions like
```
if (strncmp(argv[i], "option", strlen("option")) == 0)
```
or
```
if (strncmp(argv[i], "option", 7) == 0)
```
or
```
if (strncmp(argv[i], "option", 7) == 0 && strncmp(argv[i], "option", strlen(argv[i])) == 0)
```
or even
```
if (! (strncmp(argv[i], "option", __maxlen(7, strlen(argv[i])) == 0) )
```
In some places even `strncasecmp()` is used.

All those approaches look awfully redundant, cumbersome and unprofessional. They also lead to problems described in aforementioned ticket.

This issue is to capture the effort to clean them all up and switch to using simple `strcmp()` wherever normal C strings coming from command line arguments or console input are compared.

There are also places where `strncmp()` is used to compare IPMI strings without the null-byte terminator and with a type/length byte prefix instead. Those must be left as is (or maybe converted to using some new helper function like `ipmi_strcmp()`).

doublechiang:
I created a new branch name to reflect the changes.

### DIFF
--- a/lib/helper.c
+++ b/lib/helper.c
@@ -647,7 +647,7 @@ uint32_t str2val32(const char *str, const struct valstr *vs)
 	int i;
 
 	for (i = 0; vs[i].str; i++) {
-		if (strncasecmp(vs[i].str, str, __maxlen(str, vs[i].str)) == 0)
+		if (strcasecmp(vs[i].str, str) == 0)
 			return vs[i].val;
 	}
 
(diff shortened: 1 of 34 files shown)


AlexanderAmelkin:
This is nice, but since you've already submitted a PR with your master branch, and that PR is half-reviewed already, let's not create more. Especially taking in account that the branch name doesn't actually mean much. It is nice to have a properly named branch, but after the change is merged, it can't be seen anywhere as I usually do fast-forward (rebase) merges.

AlexanderAmelkin:
Please commit any changes you have based on review of #185 onto your master branch.

PR #208, opened 2020-06-14
Regenerate configure script in TravisCI

Chocobo1:
This would allow catching errors in build scripts.


Chocobo1:
Oops.. Never mind.

PR #221, opened 2020-08-03
Add support for BCD and 6-bit ASCII to ipmitool SDR handling

cminyard:
This has caused me grief several times.  The OpenIPMI SDR compiler will create BCD or 6-bit ASCII names per the spec, but ipmitool wouldn't handle them even though it has been in the spec since the beginning.

Add support the ipmitool for these.  This actually results in much neater code, IMHO.

The second patch switches FRU string handling over to the new string handling.  This is more important than you might imagine.  Binary handling of FRU data is broken, for one.  There is no end of data handling for FRU strings.  A carefully crafted FRU string might be able to cause an overrun in ipmitool and do bad things.

There are four semantic changes that should happen with these:
 * SDRs with BCD or 6-bit ASCII will print properly instead of printing gibberish.
 * SDR strings that are 16 bytes long will not print past the end of the string.  There were a bunch of:
      printf("xxx %s\n", sdr->xxx->id_string);
   thoughout the code, and the string in the SDR can be the full size.
 * FRU data with binary formatting will print as a hex dump instead of printing gibberish.
 * A FRU data string that went past the end of the area will no longer read past the end of the area.

I think all of these are good things.  But you can take just the SDR changes without the FRU changes, too.


cminyard:
Well, github found some minor issues, let me try again.

Issue #229, opened 2020-08-25
IPMI User Password bug.


mbap:
User password is broken via getpass.

Patch below.
```
commit 30e25cfa3e985d276a9aec16490d74aad731a15f
Author: Michael Baptist <michael.baptist@isilon.com>
Date:   Tue Aug 25 00:33:55 2020 -0500

    Fix user password matching from /dev/tty
    
    This is a quick fix for password matching. getpass was not being used
    correctly because it uses a static buffer so pointers to the getpass
    return value will all point to the same buffer. Comparing them after
    multiple calls will always compare equally.
    
    This takes the return value to getpass and copies it into a local buffer
    for comparison. There are additional improvements that could be made
    such as wiping the password buffers after use to remove passwords from
    memory faster, but I'll leave that for the current maintainers.

diff --git a/lib/ipmi_user.c b/lib/ipmi_user.c
index 2068c93672a8cc33bcb3c972c5ffa617092591ee..c9ec8d60b9fb6945c05f4c6fe2e53846db03e930 100644
--- a/lib/ipmi_user.c
+++ b/lib/ipmi_user.c
@@ -48,6 +48,10 @@
 #include <ipmitool/ipmi_strings.h>
 #include <ipmitool/bswap.h>
 
+#define USER_PW_IPMI15_LEN 16 /* IPMI 1.5 only allowed for 16 bytes */
+#define USER_PW_IPMI20_LEN 20 /* IPMI 2.0 allows for 20 bytes */
+#define USER_PW_MAX_LEN USER_PW_IPMI20_LEN
+
 
 extern int verbose;
 extern int csv_output;
@@ -477,16 +481,31 @@ ipmi_user_build_password_prompt(uint8_t user_id)
  *
  * @returns pointer to char with password
  */
-char *
-ask_password(uint8_t user_id)
+int
+ask_password(uint8_t user_id, char *buf, size_t buflen)
 {
-	const char *password_prompt =
-		ipmi_user_build_password_prompt(user_id);
-# ifdef HAVE_GETPASSPHRASE
-	return getpassphrase(password_prompt);
-# else
-	return (char*)getpass(password_prompt);
-# endif
+	const char *password_prompt;
+	char *sbuf;
+	size_t len;
+	int ret = 0;
+
+	password_prompt = ipmi_user_build_password_prompt(user_id);
+	if (buf) {
+		sbuf = getpass(password_prompt);
+		len = strnlen(sbuf, buflen + 1);
[...]

PR #243, opened 2020-09-30
ipmitool: Fix ask_password getpass usage.

mbap:
This is a quick fix for password matching. getpass was not being used
correctly because it uses a static buffer. Pointers to the getpass
return value will all point to that same static buffer. Comparing them
after multiple calls in incorrect.

This fix takes the return value to getpass and stores it into a local
buffer for comparison. There are additional improvements that could be
made such as wiping the pw buffers asap to reduce duration in memory.


mbap:
```
$ipmitool user set password 9              
Password for user 9: 
Password for user 9: 
Set User Password command successful (user 9)

$ipmitool user set password 9 
Password for user 9: 
Password for user 9: 
Set User Password command successful (user 9)

$ipmitool user set password 9 
Password for user 9: 
Password for user 9: 
Passwords do not match or are longer than 20

$ipmitool user set password 9 
Password for user 9: 
Password for user 9: 
Set User Password command successful (user 9)

$ipmitool user set password 9 
Password for user 9: 
ipmitool: password input exceeded buffer

$ipmitool user set password 9 
Password for user 9: 
Password for user 9: 
Set User Password command successful (user 9)

$ipmitool user test 9 20 
Password for user 9: 
Failure: wrong password size

$ipmitool user set password 9 
Password for user 9: 
Password for user 9: 
Set User Password command successful (user 9)

$ipmitool user test 9 20      
Password for user 9: 
Success
```

mbap:
Closes https://github.com/ipmitool/ipmitool/issues/229

lib/ipmi_user.c
```diff
@@ -529,9 +548,11 @@ int
 ipmi_user_test(struct ipmi_intf *intf, int argc, char **argv)
 {
 	/* Test */
-	char *password = NULL;
+	char *password;
+	char passbuf[USER_PW_MAX_LEN+1];
```
AlexanderAmelkin, line 552: spaces around `+` please

lib/ipmi_user.c
```diff
@@ -626,42 +648,41 @@ ipmi_user_mod(struct ipmi_intf *intf, int argc, char **argv)
...
-#define USER_PW_IPMI20_LEN 20 /* IPMI 2.0 allows for 20 bytes */
-#define USER_PW_MAX_LEN USER_PW_IPMI20_LEN
-
 int
 ipmi_user_password(struct ipmi_intf *intf, int argc, char **argv)
 {
-	char *password = NULL;
+	char *password;
+	char passbuf[USER_PW_MAX_LEN+1];
+	char tmp[USER_PW_MAX_LEN+1];
```
AlexanderAmelkin, line 656: here too

AlexanderAmelkin (changes requested):

lib/ipmi_user.c
```diff
@@ -477,16 +481,31 @@ ipmi_user_build_password_prompt(uint8_t user_id)
...
  * @returns pointer to char with password
  */
-char *
-ask_password(uint8_t user_id)
+int
+ask_password(uint8_t user_id, char *buf, size_t buflen)
 {
-	const char *password_prompt =
-		ipmi_user_build_password_prompt(user_id);
-# ifdef HAVE_GETPASSPHRASE
```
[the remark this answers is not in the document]
mbap: Correct me if I am wrong.

`getpassphrase(3)` is a Solaris only API?
`termios(3)` is recommended instead of `getpass(3)` on Linux according to the man page, however I believe that mainly to be true due to the ease of introducing bugs such as the one I was attempting to fix here. The getpass implementation just uses termios under the hood and uses a static buffer (foolishly?) rather than taking a buffer as an arg, so the code is mostly the same anyways.
`getpass` is not depricated in the BSD manpages (FreeBSD specifically), however they have their own passphrase API `readpassphrase(3)`

I decided to continue the use of getpass and remove the Solaris API to first simply the code, and also maintain some portability without if def cpp all over the place.

Will await further instruction.

lib/ipmi_user.c
mbap: Fixed.

lib/ipmi_user.c
mbap: Fixed.

lib/ipmi_user.c
```diff
@@ -477,16 +481,31 @@ ipmi_user_build_password_prompt(uint8_t user_id)
...
-# else
-	return (char*)getpass(password_prompt);
-# endif
+	const char *password_prompt;
+	char *sbuf;
+	size_t len;
+	int ret = 0;
+
+	password_prompt = ipmi_user_build_password_prompt(user_id);
+	if (buf) {
```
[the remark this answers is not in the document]
mbap: Fixed.

mbap:
Awaiting response on getpass/getpassphrase/readpassphrase discussion. Otherwise initial comments should be resolved.

lib/ipmi_user.c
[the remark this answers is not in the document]
AlexanderAmelkin: Well, although I'm not sure Solaris compatibility is still maintained otherwise, I see that Solaris is still around (11.4 was released 2 years ago and in still maintained), and some people may be still using ipmitool for it.

Anyway, if Solaris support is to be dropped, that must be done explicitly with a separate PR affecting all the sources, not just one file. So please restore the `getpassphrase()` support here unless you're sure that getpass() can work for Solaris and, for instance, will not truncate the password to 9 characters like it says [here](https://docs.oracle.com/cd/E23823_01/html/816-5168/getpass-3c.html).

lib/ipmi_user.c
```diff
@@ -477,16 +481,31 @@ ipmi_user_build_password_prompt(uint8_t user_id)
...
+	size_t len;
+	int ret = 0;
+
+	password_prompt = ipmi_user_build_password_prompt(user_id);
+	if (buf && buflen > 0) {
+		sbuf = getpass(password_prompt);
+		len = strnlen(sbuf, buflen + 1);
+		if (len < buflen) {
+			memcpy(buf, sbuf, len * sizeof(char));
+			buf[len+1] = '\0';
```
AlexanderAmelkin, line 498: Spaces around `+`, please

lib/ipmi_user.c
```diff
@@ -477,16 +481,31 @@ ipmi_user_build_password_prompt(uint8_t user_id)
...
-		ipmi_user_build_password_prompt(user_id);
-# ifdef HAVE_GETPASSPHRASE
-	return getpassphrase(password_prompt);
-# else
-	return (char*)getpass(password_prompt);
-# endif
+	const char *password_prompt;
+	char *sbuf;
+	size_t len;
+	int ret = 0;
```
AlexanderAmelkin, line 490: It's preferred to default to an error condition rather than success. A success must be explicit.

lib/ipmi_user.c
```diff
@@ -626,42 +648,41 @@ ipmi_user_mod(struct ipmi_intf *intf, int argc, char **argv)
...
 			return (-1);
 		}
-		if (strncmp(password, tmp, tmplen)) {
+		tmplen = strnlen(tmp, USER_PW_MAX_LEN + 1);
+		if (strncmp(passbuf, tmp, tmplen)) {
 			lprintf(LOG_ERR, "Passwords do not match or are "
 			                 "longer than %d", USER_PW_MAX_LEN);
 			return (-1);
 		}
+		/* XXX memcpy out passwords asap. */
```
AlexanderAmelkin, line 684: I don't exactly understand this comment here. Who and when do you want to do this?

AlexanderAmelkin (changes requested):

lib/ipmi_user.c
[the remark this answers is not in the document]
mbap: Yep that is fair. I will do my best there, I don't have a means of testing Solaris specifically, but ill reference the man page. I'll update this PR asap. Thanks.

lib/ipmi_user.c
mbap: It's more or less a warning that a without memset/bzeroing the password buffers immediately after use, the user password will be resident in memory. It's likely a low risk attack vector, since ipmitool exits shortly after invocation in this case, however on an idle system, if the virtual memory pages are not reused rather quickly, in theory the passwords could be read by someone with know how.

lib/ipmi_user.c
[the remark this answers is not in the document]
mbap: Reading the man page, getpass on Solaris only supports 9 bytes, so that makes sense why getpassphrase is needed. Fixing in my latest rev.

lib/ipmi_user.c
AlexanderAmelkin: Yes, that is clear, I'm just wondering who is the target of this comment. Who exactly needs to do this, when they are going to do this (never, I figure), and if it needs to be done, then why not do it right away?

lib/ipmi_user.c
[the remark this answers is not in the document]
AlexanderAmelkin: Thanks. Will wait for the termios version then.

lib/ipmi_user.c
```diff
@@ -477,16 +481,47 @@ ipmi_user_build_password_prompt(uint8_t user_id)
...
+	 */
+#ifdef HAVE_GETPASSPHRASE
+	password_routine = &getpassphrase;
+#else
+	password_routine = &getpass;
+#endif
+
+	password_prompt = ipmi_user_build_password_prompt(user_id);
+	if (buf && buflen > 0) {
+		sbuf = password_routine(password_prompt);
```
pstrinkle, line 507: nit: you can reduce the indentation with some early bails - I don't know what the policy is on these, but if sbuf == NULL) { log; return ret; } etc etc

lib/ipmi_user.c
```diff
@@ -550,18 +588,22 @@ ipmi_user_test(struct ipmi_intf *intf, int argc, char **argv)
...
 		}
+		password = passbuf;
 	} else {
 		password = argv[3];
 	}
-	return ipmi_user_test_password(intf,
+	ret = ipmi_user_test_password(intf,
 					 user_id,
 					 password,
 					 password_length == 20);
```
pstrinkle, line 604: Magic number - 20

lib/ipmi_user.c
```diff
@@ -626,42 +668,43 @@ ipmi_user_mod(struct ipmi_intf *intf, int argc, char **argv)
...
 		/* We need to prompt for a password */
-		char *tmp;
-		size_t tmplen;
-		password = ask_password(user_id);
-		if (!password) {
-			lprintf(LOG_ERR, "ipmitool: malloc failure");
+		memset(passbuf, 0, sizeof(passbuf));
+		memset(tmp, 0, sizeof(tmp));
+
+		if (ask_password(user_id, passbuf, sizeof(passbuf))) {
```
pstrinkle, line 692: Above you're memsetting the password buffer in all cases, but not here.

mbap:
My apologies, I haven't had cycles to get the termios work started yet. Feel free to accept/deny this PR. If accepted, please feel free to file an issue and assign me to take to completion. I'm hoping to get some time this week to work this.

mbap:
@pstrinkle Browser cache got me.. 

Will take a look at closer look at your review  asap.

lib/ipmi_user.c
mbap: I'll make a define for the IPMI 1.5 vs 2.0 pw. ( I think it was 1.5 at least.).

Wasn't sure the tolerance on churn, but easy fix.

lib/ipmi_user.c
```diff
@@ -626,42 +668,43 @@ ipmi_user_mod(struct ipmi_intf *intf, int argc, char **argv)
...
 			return (-1);
 		}
-		tmp = ask_password(user_id);
-		tmplen = strnlen(tmp, USER_PW_MAX_LEN + 1);
-		if (!tmp) {
-			lprintf(LOG_ERR, "ipmitool: malloc failure");
+		if (ask_password(user_id, tmp, sizeof(tmp))) {
```
vmauery, line 695: same here; no memset

lib/ipmi_user.c
```diff
@@ -477,16 +481,47 @@ ipmi_user_build_password_prompt(uint8_t user_id)
...
+	 * Note: getpass appears to be a trivial portable password routine,
+	 * however it is considered obsolete by most modern OSs. Unfortunately
+	 * there is not really a portable solution that was obvious so getpass
+	 * is currently to remain in use.
+	 */
+#ifdef HAVE_GETPASSPHRASE
+	password_routine = &getpassphrase;
+#else
+	password_routine = &getpass;
+#endif
```
vmauery, line 503: instead of a function pointer created with an ifdef, why not just a #define that sets it?

#ifdef HAVE_GETPASSPHRASE
#define password_routine getpassphrase
#else
#define password_routine getpass
#endif

vmauery: looks like this is moot if we are going with a termios/fgets situation anyway.

lib/ipmi_user.c
mbap: input buffer isn't touched in the failure case.

mbap:
Ok, going to test these changes this week. Took at stab at it. It is likely not perfect, but hoping it is sufficient. Careful review would be awesome specifically looking for buffer overflow or careless errors in that regard.

lib/ipmi_user.c
```diff
@@ -461,32 +470,104 @@ print_user_usage(void)
...
+			err = -1;
+			errno = ENOBUFS;
+			break;
+		}
+		*idx++ = data;
+	}
+
+	save_errno = errno;
+	*idx = '\0';
+	write(out, "\n", 1);
```
mbap, line 548: Should be sizeof(char)

lib/ipmi_user.c
```diff
@@ -461,32 +470,104 @@ print_user_usage(void)
...
+
+	if (tty) {
+		while ((tcsetattr(in, TCSAFLUSH | TCSASOFT, &prev) == -1) && errno == EINTR) {
+			continue;
+		}
+		close(in);
+	}
+
+	if (err) {
```
mbap, line 557: if (err == -1)

lib/ipmi_user.c
```diff
@@ -461,32 +470,104 @@ print_user_usage(void)
...
+		if (len < buflen) {
+			memcpy(buf, ibuf, len * sizeof(char));
+			buf[len + 1] = '\0';
+			ret = 0;
+		} else {
+			lprintf(LOG_ERR, "ipmitool: password input exceeded buffer");
+		}
+	}
+	memset(ibuf, 0, sizeof(ibuf));
+	return 0;
```
mbap, line 570: return ret;

lib/ipmi_user.c
```diff
@@ -541,27 +625,33 @@ ipmi_user_test(struct ipmi_intf *intf, int argc, char **argv)
...
 	}
 	if (argc == 3) {
 		/* We need to prompt for a password */
-		password = ask_password(user_id);
-		if (!password) {
-			lprintf(LOG_ERR, "ipmitool: malloc failure");
+		memset(passbuf, 0, sizeof(passbuf));
+		if (ask_password(user_id, passbuf, sizeof(passbuf))) {
+			lprintf(LOG_ERR, "ipmitool: password failure");
+			memset(passbuf, 0, sizeof(passbuf));
```
mbap, line 642: unneeded memset.

mbap, line 642: unneeded memset

mbap:
Not totally sure whats going on with CI, can't seem to make catalina happy, however I'm building locally just fine on 10.15.7.

mbap:
## IPMITool user password testing.

### Configure test user.
```
$./ipmitool  user set name 9 ""
$./ipmitool  user disable 9
$./ipmitool  user set name 9 "usertest"
$./ipmitool  user enable 9
$./ipmitool  user list
ID  Name	     Callin  Link Auth	IPMI Msg   Channel Priv Limit
9   usertest      true    false      false      NO ACCESS
```

### Test non matching passwords.
```
$./ipmitool  user set password 9
Password for user 9:
Password for user 9:
Passwords do not match or are longer than 20
```

### Test password greater than 20 chars.
```
$./ipmitool  user set password 9
Password for user 9:
ipmitool: password input error: No buffer space available
```

### Test setting empty password ("")
```
$./ipmitool  user set password 9
Password for user 9:
Password for user 9:
Set User Password command successful (user 9)
```

### Test set matching password (testpw)
```
$./ipmitool  user set password 9
Password for user 9:
Password for user 9:
Set User Password command successful (user 9)
```

### Test correct password.
```
$./ipmitool  user test 9 16
Password for user 9:
Success
```

## Test incorrect Password
```
$./ipmitool  user test 9 16
Password for user 9:
Failure: password incorrect
```

mbap:
Tested on Fedora 33 over the lanplus interface to a Dell PowerEdge r640.

mbap:
Sorry for the long delay.

AlexanderAmelkin:
> Sorry for the long delay.

Thank you for your time. I will take a look at it as soon as I have some spare time too.

filipjares:
Hello everybody, please forgive my question if it is inappropriate.

I encountered the bug documented in #229 independently and created a simple patch myself. Then I found this pull request.

I can see that this pull request has more ambitions than merely fixing the bug itself and it also intends to improve the code.

I can see there are failed checks in this pull request, but their logs are not available any more.
Can I help with this pull request? Or is it just waiting for review? Does it make sense to create an alternative/simpler pull request or would that just create an unnecessary overhead?

AlexanderAmelkin:
@filipjares, it looks to me that the failed builds are caused by some changes in GitHub Actions infrastructure, as they just started failing on new PRs without me merging anything that failed them. If you're willing to investigate that, you're more than welcome.

lib/ipmi_user.c
mbap: Marking resolved since no feedback was returned.

mbap:
Anything I can do to help move this along?

PR #247, opened 2020-10-19
channel: fix the error of converting "alert on/off" to raw command

ryanckfang:
From IPMI Spec, Chapter 22.22 Set Channel Access Command
Table 22, Set Channel Access Command

Byte#2, Bit#5 is "PEF Alerting Enable/Disable"
And the bit value:
    0b = enable PEF Alerting
    1b = disable PEF Alerting on this channel

In current code, alert "on" set Bit#5 to 1 and alert "off" set Bit#5 to
0, it's straightforward but just opposite of IPMI spec bit definition.


lib/ipmi_channel.c
```diff
@@ -182,7 +182,7 @@ _ipmi_set_channel_access(struct ipmi_intf *intf,
 	memset(&data, 0, sizeof(data));
 	data[0] = channel_access.channel & 0x0F;
 	data[1] = (access_option << 6);
-	if (channel_access.alerting) {
+	if (!channel_access.alerting) {
```
AlexanderAmelkin, line 185: This change is completely wrong and just breaks everything. Please see carefully how and where this structure field is used.

AlexanderAmelkin: The bug is, nonetheless, present indeed. It needs to be fixed elsewhere though. I'll do it.

AlexanderAmelkin:
Please confirm that my version of the fix works for you. I don't have any hardware that supports this feature to test on.

ryanckfang:
> Please confirm that my version of the fix works for you. I don't have any hardware that supports this feature to test on.

After tested, your fix works in my hardware platform.

PR #252, opened 2020-11-12
sel: Remove redundant "Reserve SEL"

louiswpf:
There is no need to reserve SEL when doing "sel list" and "sel get"
since we don't do partial get in both commands, and we didn't even use
the reservation ID returned by "Reserve SEL".

Signed-off-by: Bing-Hua Wang <binghuawang@ami.com>


louiswpf:
I've tackled the macOS build breakage in #255.

louiswpf:
@AlexanderAmelkin Please review.

AlexanderAmelkin (approved):
Looks like you're right and reservation is indeed not needed in those functions.

PR #255, opened 2020-12-04
Clean up headers and fix the macOS build

louiswpf:
The macOS build breakage is due to missing includes. However after adding the
missing includes, the build still fails. It turns out there are header guard
name collisions between freeipmi and ipmitool.

In this PR, we fix the header guard name collisions first, do some small
cleanups while we're at it, and finally fix the macOS build.


include/ipmitool/ipmi_sdr.h
```diff
@@ -30,8 +30,8 @@
  * EVEN IF SUN HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH DAMAGES.
  */
 
-#ifndef IPMI_SDR_H
-#define IPMI_SDR_H
+#ifndef IPMITOOL_IPMI_SDR_H
+#define IPMITOOL_IPMI_SDR_H
```
AlexanderAmelkin, line 34: This is a great finding, but if Wikipedia is to be trusted, `#pragma once` is nowadays supported by all compilers on all systems, so it would be better to switch to it as long as we're changing every include file anyway.

See https://en.wikipedia.org/wiki/Pragma_once#Portability

AlexanderAmelkin (changes requested):

include/ipmitool/ipmi_sdr.h
louiswpf: Nice suggestion. Addressed.

louiswpf:
@AlexanderAmelkin I've also fixed the Windows CI breakage and renamed the topic branch. This PR somehow is closed automatically by GitHub due to branch renaming. Thus I have to open another PR #284. Please take a look.

PR #261, opened 2021-01-25
Fix ioctl(TIOCFLUSH) call

sthibaul:
BSD 4.2 introduced a parameter to the TIOCFLUSH call to specify which
directions should be flushed. The TIOCFLUSH ioctl thus expects a
proper argument, otherwise it would crash.

Signed-off-by: Samuel Thibault <samuel.thibault@ens-lyon.org>


src/plugins/serial/serial_basic.c
```diff
@@ -328,7 +328,8 @@ serial_bm_flush(struct ipmi_intf * intf)
 #if defined(TCFLSH)
     return ioctl(intf->fd, TCFLSH, TCIOFLUSH);
 #elif defined(TIOCFLUSH)
-    return ioctl(intf->fd, TIOCFLUSH);
+    int zero = 0;
+    return ioctl(intf->fd, TIOCFLUSH, &zero);
```
AlexanderAmelkin, line 332: I assume it worked before BSD 4.2 ? Looks like this patch breaks compatibility. Probably, you need to add some additional feature macro checking, don't you?

AlexanderAmelkin (changes requested):

src/plugins/serial/serial_basic.c
sthibaul: > I assume it worked before BSD 4.2 ? 

BSD 4.1 and before would not read the 3rd parameter at all indeed, and thus be fine with no 3rd parameter be given.

> Looks like this patch breaks compatibility.

No. the ioctl 3rd parameter is optional, so calling with it or without it does not change the processing of the first two parameters, so BSD 4.1 would be completely fine with the 3 parameters.

> Probably, you need to add some additional feature macro checking, don't you?

Just for information, BSD 4.2 was released in 1983. I thus don't think it's worth discussing more about such a compatibility concern. I mentioned BSD 4.1 only to explain why you might find some source code on the net that does not include a 3rd parameter for TIOCFLUSH. (though in the entire debian source code I only found ipmitool that does such a thing).

PR #263, opened 2021-01-27
Fix double bridge detection

patrislav1:
Consider a MicroTCA system and using ipmitool to talk through the MCH to a AMC module directly:
```
ipmitool -B 0 -T 0x82 -b 7 -t <IPMB_ADDR> -H <MCH_IPADDR> <COMMAND>
```

[ipmi_intf_get_max_request_data_size()](https://github.com/ipmitool/ipmitool/blob/050f5da63106e89ba11da3e5bac3f1e76209a3ef/src/plugins/ipmi_intf.c#L501) and [ipmi_intf_get_max_response_data_size()](https://github.com/ipmitool/ipmitool/blob/050f5da63106e89ba11da3e5bac3f1e76209a3ef/src/plugins/ipmi_intf.c#L558) check if the IPMI access uses double bridging and reserve 8 bytes for extra overhead (forwarding the message from first bridge over IPMB-0 to second bridge).

The [existing implementation](https://github.com/ipmitool/ipmitool/blob/050f5da63106e89ba11da3e5bac3f1e76209a3ef/src/plugins/ipmi_intf.c#L543) determines double bridging mode from `transit_addr` (-T option) being not equal to `target_addr` (-t option).
If the AMC is in a slot where the address on IPMB-L bus matches the address of the carrier manager on IPMB-0 bus (e.g. `-T 0x82 -t 0x82` with AMC in 9th slot), then the double bridging is not detected, so that the overhead is not reserved and the resulting IPMB packet gets too large - especially with commands like `HPMFWUPG_UPLOAD_FIRMWARE_BLOCK` where ipmitool maxes out the payload. (See screenshot: 40 bytes are sent over IPMB-0, which is out of spec, AFAIK).

This patch fixes the double bridge detection by not only comparing the addresses, but also the channels of the transit and target. Double bridge is detected if either is different.

In the second commit I also refactored the bridging level detection to a dedicated function. I noticed that the detection of forwarding and bridging is re-implemented in a lot of places - I would be willing to refactor some of them as well, using `ipmi_intf_get_bridging_level()`, if it gets a chance of being merged upstream.

ipmi-dbridge-bug


patrislav1:
Any chance of getting this bugfix merged?

AlexanderAmelkin (approved):

AlexanderAmelkin:
Thanks for this awesome fix!

AlexanderAmelkin:
And thanks for the awesome PR description too! I wish everyone did it like that!

PR #264, opened 2021-01-28
Update helper.c

jmoellers:
The third fd is STDERR_FILENO.

### DIFF
--- a/lib/helper.c
+++ b/lib/helper.c
@@ -935,7 +935,7 @@ ipmi_start_daemon(struct ipmi_intf *intf)
 		exit(1);
 	}
 	ret = dup(fd);
-	if (ret != STDOUT_FILENO) {
+	if (ret != STDERR_FILENO) {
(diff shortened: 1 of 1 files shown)


lib/helper.c
```diff
@@ -935,7 +935,7 @@ ipmi_start_daemon(struct ipmi_intf *intf)
 		exit(1);
 	}
 	ret = dup(fd);
-	if (ret != STDOUT_FILENO) {
+	if (ret != STDERR_FILENO) {
```
AlexanderAmelkin, line 938: Looks like nobody has ever used this code because you're the first to notice this.
Thanks.

AlexanderAmelkin (approved):

lib/helper.c
AlexanderAmelkin: Please change the commit message to
```
helper: Fix stderr file descriptor

The third file descriptor is STDERR_FILENO, not STDOUT_FILENO
```

AlexanderAmelkin (changes requested):

PR #265, opened 2021-02-01
Fixed compiler warning.

shitalkumar-gandhi:
This patch has been added to fix "error: comparison between signed
and unsigned integer expressions [-Werror=sign-compare]"

Signed-off-by: Shitalkumar Gandhi <shitalkumar.gandhi@seagate.com>

### DIFF
--- a/lib/ipmi_tsol.c
+++ b/lib/ipmi_tsol.c
@@ -376,7 +376,7 @@ ipmi_tsol_main(struct ipmi_intf *intf, int argc, char **argv)
 	char out_buff[IPMI_BUF_SIZE * 8];
 	char buff[IPMI_BUF_SIZE + 4];
 	int fd_socket, result, i;
-	int out_buff_fill, in_buff_fill;
+	size_t out_buff_fill, in_buff_fill;
 	int ip1, ip2, ip3, ip4;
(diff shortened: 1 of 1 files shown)


lib/ipmi_tsol.c
```diff
@@ -534,7 +534,7 @@ ipmi_tsol_main(struct ipmi_intf *intf, int argc, char **argv)
 
 		if ((fds[0].revents & POLLIN) && (sizeof(out_buff) > out_buff_fill)) {
 			socklen_t sin_len = sizeof(sin);
-			int buff_size = sizeof(buff);
+			unsigned short int buff_size = sizeof(buff);
```
AlexanderAmelkin, line 537: The type of `sizeof()` is `size_t`, please use it. Please remember that there are different architectures too.

AlexanderAmelkin (changes requested):

lib/ipmi_tsol.c
shitalkumar-gandhi: > The type of `sizeof()` is `size_t`, please use it. Please remember that there are different architectures too.

Thanks, done.

AlexanderAmelkin (approved):

PR #266, opened 2021-02-03
fix unit

jason63716:
According to ipmi2.0 protocol,page552,"Sensor  Units 1" field, Modifier Unit is decided by "Sensor Units 3"， so place must be unit_desc[modifier], other than unit_desc[base];


AlexanderAmelkin (changes requested):
Thanks.
Please change the commit message to:
```
sdr: Fix modifier unit

Previously instead of the modifier unit, the base unit was printed. 
```

Please also use `-s` flag with `git commit --amend` when updating the message.

AlexanderAmelkin (approved):

PR #278, opened 2021-03-29
feat: Add hpeoem lan command to configure HPE iLO NIC selection

jiegec:
Add supports for HPE iLO NIC selection. Example usage:

```shell
> sudo ./src/ipmitool hpeoem lan get 1
NIC selection: shared
Reset needed: no
Shared NIC selection: ALOM
Shared NIC port: NIC port 1
Shared NIC support: ALOM
> sudo ./src/ipmitool hpeoem lan set 1 dedicated
Set NIC selection mode to dedicated
> sudo ./src/ipmitool hpeoem lan get 1
NIC selection: dedicated
Reset needed: yes
```

It uses the command documented in: https://support.hpe.com/hpesc/public/docDisplay?docId=c04530505&docLocale=en_US


lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,428 @@
...
+            "      returns the current active NIC.");
+    lprintf(LOG_NOTICE,
+            "");
+}
+
+static int
+ipmi_hpeoem_lan_get_nic_selection(struct ipmi_intf *intf, int chan)
+{
+    struct ipmi_rs *rsp;
+    struct ipmi_rq req;
```
AlexanderAmelkin, line 208: Please either initialize to `{}` or `memset()` it to zeroes later before first use.

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,428 @@
...
+    uint8_t input_length = 4;
```
AlexanderAmelkin, line 209: Please avoid arbitrary magic numbers. This is actually `sizeof(msg_data)` and you don't need a variable for that.

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,428 @@
...
+    uint8_t msg_data[4];
+    uint8_t nic_selection = -1;
+    uint8_t shared_nic_selection = -1;
+    uint8_t shared_nic_port = -1;
+    char *nic_selection_str = "Unknown";
+    char *shared_nic_selection_str = "Unknown";
+    char *shared_nic_port_str = "Unknown";
+    uint8_t reset_needed = 0;
```
AlexanderAmelkin, line 217: better use `bool` type from `#include <stdbool.h>` for boolean values.

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,428 @@
...
+    "NIC port 1",
+    "NIC port 2",
+};
+
+#define INVALID -1
+#define DEDICATED 0
+#define SHARED_WITH_LOM1 1
+#define SHARED_WITH_LOM2 2
+#define SHARED_WITH_ALOM1 3
+#define SHARED_WITH_ALOM2 4
```
AlexanderAmelkin, line 76: I'd suggest using an `enum` as it is better for debugging. Enums are displayed as names in GDB, macros aren't.

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,428 @@
...
+    rsp = intf->sendrecv(intf, &req);
+    if (!rsp)
+    {
+        lprintf(LOG_ERR, "Error in getting nic selection");
+        return -1;
+    }
+    else if (rsp->ccode)
+    {
+        lprintf(LOG_ERR, "Error in getting nic selection (%s)",
+                val2str(rsp->ccode, completion_code_vals));
```
AlexanderAmelkin, line 239: `IPMI_LAN_GET_CONFIG` and `IPMI_LAN_SET_CONFIG` can return command-specific codes that aren't covered by `completion_code_vals`. Check `get_lan_cc_vals` in `lib/ipmi_lanp.c`.

I think that instead of reimplementing LAN parameter setting and getting here, you could just reuse the existing `get_lan_param()` and `set_lan_param_nowait()` functions (you'll need to make those public).

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,428 @@
...
+static int
+ipmi_hpeoem_lan_get_nic_selection(struct ipmi_intf *intf, int chan)
+{
+    struct ipmi_rs *rsp;
+    struct ipmi_rq req;
+    uint8_t input_length = 4;
+    uint8_t msg_data[4];
+    uint8_t nic_selection = -1;
+    uint8_t shared_nic_selection = -1;
+    uint8_t shared_nic_port = -1;
```
AlexanderAmelkin, line 213: Don't all those `-1` values stand for `INVALID` from line 71 above? Please use that macro/enum instead. If you go there with enum (preferred), then please also update the variable types here.

include/ipmitool/ipmi_hpeoem.h
```diff
@@ -0,0 +1,54 @@
...
+
+#define IPMI_ILO_SHARED_NIC_SELECTION 197
+#define IPMI_ILO_NETWORK_CONFIG_NIC_SELECTION 224
+
+#define ILO_SELECT_NIC_DEDICATED 0
+#define ILO_SELECT_NIC_SHARED 1
+
+// #224 data 4
+#define ILO_CONFIGURATION_STATUS_MASK 1
+#define ILO_CONFIGURATION_STATUS_RESET_NEEDED 1
```
AlexanderAmelkin, line 46: This is never used anywhere

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    nic_selection = rsp->data[3];
+    if (nic_selection <= 1)
+    {
+        nic_selection_str = ILONICSelectionMode[nic_selection];
+    }
+    printf("NIC selection: %s\n", nic_selection_str);
+
+    reset_needed = (rsp->data[4] & ILO_CONFIGURATION_STATUS_MASK) == ILO_CONFIGURATION_STATUS_MASK;
```
AlexanderAmelkin, line 250: I believe you indented to use `ILO_CONFIGURATION_STATUS_RESET_NEEDED` on the right of `==` here.
Nonetheless, I think it's not worth it anyway. For a single-bit mask you could simply use this:
`reset_needed = !!(rsp->data[4] & ILO_CONFIGURATION_STATUS_MASK);`

Also, please ensure that your code fits in 80 characters width unless absolutely necessary to extend.

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,428 @@
...
+    return INVALID;
+}
+
+static int
+ipmi_hpeoem_lan_set_nic_selection(struct ipmi_intf *intf, int channel, int nic_selection)
+{
+    struct ipmi_rs *rsp;
+    struct ipmi_rq req;
+    uint8_t input_length = 0;
+    uint8_t msg_data[10];
```
AlexanderAmelkin, line 347: Why `10` if the max you use is `6` ? It looks like `6` is actually the maximum defined for command 224 by iLO manual on page 141.

All in all, please avoid using magic numbers like `10` or `6`. Please use enums or macros with descriptive names.

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,428 @@
...
+    }
+    return INVALID;
+}
+
+static int
+ipmi_hpeoem_lan_set_nic_selection(struct ipmi_intf *intf, int channel, int nic_selection)
+{
+    struct ipmi_rs *rsp;
+    struct ipmi_rq req;
+    uint8_t input_length = 0;
```
AlexanderAmelkin, line 346: Why set to `0` if then immediately overridden with `6` ?

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,428 @@
...
+    uint8_t msg_data[10];
+
+    input_length = 6;
+    msg_data[0] = channel;
+    msg_data[1] = IPMI_ILO_NETWORK_CONFIG_NIC_SELECTION;
```
AlexanderAmelkin, line 351: Please think about reusing `set_lan_param_nowait()`.

AlexanderAmelkin (changes requested):

include/ipmitool/ipmi_hpeoem.h
```diff
@@ -0,0 +1,54 @@
+/****************************************************************************
+Copyright (c) 2021, Jiajie Chen
```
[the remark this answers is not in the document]
jiegec: I am not affiliated with HPE.

jiegec:
Thanks for thorough review from @AlexanderAmelkin. I will fix them ASAP.

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,430 @@
...
+    }
+    else if (argv[current_arg] && !strcmp(argv[current_arg], "lom2"))
+    {
+        return OEM_HPE_LAN_NIC_SHARED_WITH_LOM2;
+    }
+    else if (argv[current_arg] && !strcmp(argv[current_arg], "alom1"))
+    {
+        return OEM_HPE_LAN_NIC_SHARED_WITH_ALOM1;
+    }
+    else if (argv[current_arg] && !strcmp(argv[current_arg], "alom2"))
```
AlexanderAmelkin, line 335: It looks a lot like all those `if`/`else if` could be reduced to a simple dictionary and a loop:

```c
#define OEM_HPE_MAX_NIC_SIZE 10 /* the length of word 'dedicated' plus a null-byte */
struct elem {
    char type[OEM_HPE_MAX_NIC_SIZE];
    int rc;
    bool advance;
} dictionary = {
        { "dedicated", OEM_HPE_LAN_NIC_DEDICATED, false },
        { "shared", OEM_HPE_LAN_NIC_INVALID, true },
        { "with", OEM_HPE_LAN_NIC_INVALID, true },
        { "lom1", OEM_HPE_LAN_NIC_SHARED_WITH_LOM1, false },
        /*...*/
        { NULL, 0, false }
    };
    for (int i = 0; dictionary[i].name; ++i) {
        if (argv[0] && !strcmp(argv[0], dicttionary[i].name)) {
            return dictionary[i].rc;
        }
        argv += dictionary[i].advance ? 1 : 0;
    }
}
```
or something like that. Separate the code from the data.

lib/ipmi_hpeoem.c
```diff
@@ -0,0 +1,430 @@
...
+        }
+        else if (rsp->ccode)
+        {
+            lprintf(LOG_ERR, "Error in setting nic selection (%s)",
+                    val2str(rsp->ccode, completion_code_vals));
+            return -1;
+        }
+    }
+    return 0;
+}
```
AlexanderAmelkin, line 430: Newline is missing

AlexanderAmelkin (changes requested):

AlexanderAmelkin:
Also to obey the 50/72 rule, please change the commit message to:
```
oem: Add hpeoem lan commands

Add commands to configure HPE iLO NIC selection

Signed-off-by: Jiajie Chen <your@email.address>
```

AlexanderAmelkin:
Please also fix the issues found by Codacy

jiegec:
I got caught up in things recently, will continue to work on this later.

PR #284, opened 2021-04-08
ci: Fix macOS and Windows build

louiswpf:
The macOS build breakage is due to missing includes. However after adding the
missing includes, the build still fails. It turns out there are header guard
name collisions between freeipmi and ipmitool.

In this PR, we fix the header guard name collisions first, do some small
cleanups while we're at it, and finally fix the macOS build.

The Windows build breakage is due to GitHub Actions updates.
We also fix it in this PR.


AlexanderAmelkin (approved):
Thank you soooo much for this PR and all the thorough analysis that you've conducted and put into commit messages. I wish every PR was done so well.

Issue #289, opened 2021-05-02
-I lanplus hpm check command prints garbage characters


PengWang-SMARTM:
ipmitool prints the result of HPM.1 Get Component Properties (Description string) with unexpected garbage characters if it's over the -I lanplus interface.

This happens because:
1. lanplus ipmi_lan_poll_single() doesn't clear the rest of data buffer like lan ipmi_lan_poll_recv().
2. ipmi_hpmfwupg.c HpmfwupgGetComponentProperties() set the end of the string buffer, descString, to be '\0' rather than the end of actual string.

Fixing either one will eliminate the issue. I'd suggest getting both fixed.

PengWang-SMARTM:
I understand that these two file changes can be split into two different issues/branches. It shouldn't be a huge problem because both changes are simple and small enough.

Some explanations:
1. lanplus.c ipmi_lan_poll_single(), I don't know why the three lines were added onto the 1.8.18 code. I did not see a point as these variables not used anymore after that line. Therefore, I removed them.

I did not review other interface handlers and verify if they are ok or not.

2. ipmi_hpmfwupg.c HpmfwupgGetComponentProperties(). This change helps those interfaces that doesn't clear the extra data buffer. However, it will also hide the interface handler problem.

3. I initially believed the HpmfwupgGetComponentProperties() should also help the OEM properties. Then I reviewed the code. I noticed HPMFWUPG_OEM_LENGTH is defined as only 4 (ipmi_hpmfwupg.h). I guess this is specific for the manufacture who requested this feature. However, HPM.1 specification says the length is defined by the OEM ad identified by the Manufacturer ID in Get Device ID command. IPMITool, as a generic tool, should not limit this size to be 4. As I don't know the history of this part, and I don't use this OEM property, I'll leave it as is until someone complains.

PengWang-SMARTM:
I just reviewed sendrecv functions of other interfaces (1.8.18 source code).

It appears to me only the open.c and lan.c are correct with different implementations. open.c sets only the next buffer byte to be '\0' while lan.c clears entire rest of the data buffer. lan.c method is slightly more secure, but I don't feel it's necessary.

I believe, serial_terminal.c, serial_basic.c, free.c, lipmi.c, and usb.c all have the same potential problem that the command handlers, like current HpmfwupgGetComponentProperties() may accidentally access the data bytes they shouldn't.

[...]

PR #290, opened 2021-05-02
-I lanplus hpm check: fix garbage print

PengWang-SMARTM:
Two changes included in this branch:
1. zero the extra data buffer that are not containing any IPMI message
data.
2. copy only the HPM.1 property data to the pCtx response data buffer
based on the data length rather than the allowed max property size.

Resolves ipmitool/ipmitool#289

Signed-off-by: peng.wang@smartm.com <peng.wang@smartm.com>

### DIFF
--- a/lib/ipmi_hpmfwupg.c
+++ b/lib/ipmi_hpmfwupg.c
@@ -1572,9 +1572,11 @@ HpmfwupgGetComponentProperties(struct ipmi_intf *intf,
 				val2str(rsp->ccode, completion_code_vals));
 		return HPMFWUPG_ERROR;
 	}
+
+	memcpy(&pCtx->resp, rsp->data, rsp->data_len);
+	((uint8_t*)&pCtx->resp)[rsp->data_len] = '\0';
 	switch (pCtx->req.selector) {
 	case HPMFWUPG_COMP_GEN_PROPERTIES:
-		memcpy(&pCtx->resp, rsp->data, sizeof(struct HpmfwupgGetGeneralPropResp));
 		if (verbose) {
 			lprintf(LOG_NOTICE, "GENERAL PROPERTIES");
 			lprintf(LOG_NOTICE, "-------------------------------");
@@ -1591,8 +1593,6 @@ HpmfwupgGetComponentProperties(struct ipmi_intf *intf,
 		}
 		break;
 	case HPMFWUPG_COMP_CURRENT_VERSION:
-		memcpy(&pCtx->resp, rsp->data,
-				sizeof(struct HpmfwupgGetCurrentVersionResp));
 		if (verbose) {
 			lprintf(LOG_NOTICE, "Current Version: ");
 			lprintf(LOG_NOTICE, " Major: %d",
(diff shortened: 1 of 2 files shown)


lib/ipmi_hpmfwupg.c
```diff
@@ -1572,9 +1572,11 @@ HpmfwupgGetComponentProperties(struct ipmi_intf *intf,
 				val2str(rsp->ccode, completion_code_vals));
 		return HPMFWUPG_ERROR;
 	}
+
+	memcpy(&pCtx->resp, rsp->data, rsp->data_len);
+	((uint8_t*)&pCtx->resp)[rsp->data_len] = '\0';
```
AlexanderAmelkin, line 1577: Generally `rsp->data` is not a string, you can't safely terminate it like this.
Instead you need a `memset(&pCtx->resp, 0, sizeof(pCtx->resp))` before `memcpy()`.

Also this `memcpy(..., rsp->data_len)` is a security vulnerability. You should never copy blindly any size that you got from outer world. Please use `__min(rsp->data_len, sizeof(pCtx->resp))`.

AlexanderAmelkin: Also, maybe this `memset()` should go into `HpmfwupgSendCmd()` instead? Or maybe even deeper, somewhere into the interfaces `sendrecv()` calls. I don't think any of the `sendrecv()` calls should return any garbage in the `rsp`.

Please investigate this further.

src/plugins/lanplus/lanplus.c
```diff
@@ -785,9 +785,7 @@ ipmi_lan_poll_single(struct ipmi_intf * intf)
 			if (extra_data_length > 0) {
 				rsp->data_len = extra_data_length;
 				memmove(rsp->data, rsp->data + offset, extra_data_length);
-				offset = 0;
-				payload_start = 0;
-				payload_size = extra_data_length;
+				memset(rsp->data + rsp->data_len, 0, IPMI_BUF_SIZE - rsp->data_len);
```
AlexanderAmelkin, line 788: First of all, please use `sizeof(rsp->data)` instead of `IPMI_BUF_SIZE`. Never copy the size of any object from it's definition, always use `sizeof()`. Then whoever updates the definition later won't need to also update a million of other places.

Second, it looks like this change is not correct. It may need to go before `memmove()`. Or maybe it needs to go inside `read_ipmi_response()` instead or again into `sendrecv()` ?

AlexanderAmelkin (changes requested):

src/plugins/lanplus/lanplus.c
PengWang-SMARTM: I agree sizeof() is better than IPMI_BUF_SIZE. The only reason I used IPMI_BUF_SIZE is to be consistent with other drivers (e.g. Line 655 of src/lan/lan.c, end of function ipmi_lan_poll_recv()).  Please confirm you do want this to be sizeof() and I'll update the change. I don't plan to update other drivers which do not have issues.

memset() position is correct since it sets only the rest of the buffer to \0, which is actually doing the function as you mentioned in your comments for Line 1577 of lib/ipmi_hpmfwupg.c above. Moving it before memmove() will corrupt the received data before moving it to the beginning of the buffer. Please refer to the same code I just mentioned.

lib/ipmi_hpmfwupg.c
PengWang-SMARTM: Well, I agree consolidating memcpy() like this change has a potential security issue. I'll revert the changes of memcpy() and add the __min() to each memcpy() function call.

Ideally, memset() should be in sendrecv(). I agree that sendrecv() calls should  not return any garbage. However, unfortunately, as you can see my changes for the src/plugins/lanplus.c, some of interface drivers do return garbage. I mentioned them in the Issue 289. Since I don't plan to change and test all these drivers, add a memset() here will help when other interface drivers are used.

Since I've been steered away to a totally different area, I'll be very slow on this.

src/plugins/lanplus/lanplus.c
AlexanderAmelkin: I do want it to be `sizeof()`. There is plenty of less-than-optimal legacy code. Unless explicitly stated in the Coding Style, please don't blindly follow the pre-existing patterns.Thank you.

Issue #291, opened 2021-05-02
Add LAN Parameter IP Source "other" Support


PengWang-SMARTM:
HPM.3 needs the IP Source to be 4, "address obtained by BMC running other address assignment protocol".

I suggest using "other" as the name of this option, so the set command will look like "lan set x ipsrc other".

PR #292, opened 2021-05-02
lan: add ipsrc other support

PengWang-SMARTM:
LAN Parameter IP Source option 4 (other address assignment protocol) is
required by HPM.3 specification. Adding this option may help those who
are using HPM.3.

Resolves ipmitool/ipmitool#291

Signed-off-by: peng.wang@smartm.com <peng.wang@smartm.com>

### DIFF
--- a/lib/ipmi_lanp.c
+++ b/lib/ipmi_lanp.c
@@ -721,8 +721,11 @@ ipmi_lan_print(struct ipmi_intf * intf, uint8_t chan)
 		case 3:
 			printf("BIOS Assigned Address\n");
 			break;
+		case 4:
+			printf("Other Address Assignment Protocol\n");
+			break;
 		default:
-			printf("Other\n");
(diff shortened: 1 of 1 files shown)


lib/ipmi_lanp.c
```diff
@@ -2429,6 +2436,8 @@ print_lan_set_ipsrc_usage(void)
 "  dhcp   = address obtained by BMC running DHCP");
 	lprintf(LOG_NOTICE,
 "  bios   = address loaded by BIOS or system software");
+	lprintf(LOG_NOTICE,
+"  other  = address obtained via other protocol, such as HPM.3");
```
AlexanderAmelkin, line 2440: This must also be added to `doc/ipmitool.1.in`

AlexanderAmelkin (changes requested):

PR #314, opened 2021-09-15
zero initialize the recv structure on the stack

pstrinkle:
This was caught via a sanitized build where the recv was used with stack data in some cases.

### DIFF
--- a/src/plugins/open/open.c
+++ b/src/plugins/open/open.c
@@ -170,7 +170,7 @@ static
 struct ipmi_rs *
 ipmi_openipmi_send_cmd(struct ipmi_intf *intf, struct ipmi_rq *req)
 {
-	struct ipmi_recv recv;
+	struct ipmi_recv recv = {};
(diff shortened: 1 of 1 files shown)


src/plugins/open/open.c
```diff
@@ -170,7 +170,7 @@ static
 struct ipmi_rs *
 ipmi_openipmi_send_cmd(struct ipmi_intf *intf, struct ipmi_rq *req)
 {
-	struct ipmi_recv recv;
+	struct ipmi_recv recv = {};
```
AlexanderAmelkin, line 173: Good catch.
I see the same problem in `ipmievd.c:openipmi_read()`, would be nice to fix it there too.

AlexanderAmelkin (changes requested):

pstrinkle:
I submitted the separate pull request to handle the other file.  Can we merge this?

AlexanderAmelkin (approved):

PR #319, opened 2021-11-02
ipmi_sdr: Add support to print firmware change events

leiyu-bytedance:
The function ipmi_sdr_print_discrete_state() skips printing the
description for events that the data is not 0xFF.
Add "Firmware or software change detected/success" events with data 0xFF
so that ipmitool sdr could print the description.

Tested:

    ipmitool sdr elist | grep BIOS_Upgrade
    BIOS_Upgrade     | 3Eh | ok  |  7.22 | Firmware or software change success

Signed-off-by: Johny Chang <johny.chang@nettrix.com.cn>

### DIFF
--- a/include/ipmitool/ipmi_sel.h
+++ b/include/ipmitool/ipmi_sel.h
@@ -527,6 +527,7 @@ static const struct ipmi_event_sensor_types sensor_specific_event_types[] = {
     { 0x2b, 0x01, 0x15, "Firmware or software change detected, board/FRU replace new ver" },
     { 0x2b, 0x01, 0x16, "Firmware or software change detected, board/FRU replace old ver" },
     { 0x2b, 0x01, 0x17, "Firmware or software change detected, board/FRU HW conf change" },
+    { 0x2b, 0x01, 0xff, "Firmware or software change detected" },
     { 0x2b, 0x02, 0xff, "Hardware incompatibility detected" },
     { 0x2b, 0x03, 0xff, "Firmware or software incompatibility detected" },
     { 0x2b, 0x04, 0xff, "Invalid or unsupported hardware version" },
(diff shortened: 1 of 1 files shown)


include/ipmitool/ipmi_sel.h
```diff
@@ -527,6 +527,7 @@ static const struct ipmi_event_sensor_types sensor_specific_event_types[] = {
     { 0x2b, 0x01, 0x15, "Firmware or software change detected, board/FRU replace new ver" },
     { 0x2b, 0x01, 0x16, "Firmware or software change detected, board/FRU replace old ver" },
     { 0x2b, 0x01, 0x17, "Firmware or software change detected, board/FRU HW conf change" },
+    { 0x2b, 0x01, 0xff, "Firmware or software change detected" },
```
AlexanderAmelkin, line 530: This `0xff` value doesn't seem to be valid as per IPMI 2.0 specification.
The `unspecified` Event Data 2 byte value must be `0x00` (that is listed above), not `0xff`.
I may agree that `ipmitool` shouldn't silently ignore incorrect values, but then I'm curious why only `0xff` gets the honor of being handled among all other out-of-spec values.

I'd propose to alter the actual code, not just the dictionary, to print something like `Invalid sensor event data: XX XX XX` for cases like this.

AlexanderAmelkin (changes requested):

include/ipmitool/ipmi_sel.h
Journey-x: The unspecified Event Data 2 byte value must be 0xff.  Please refer to IPMI 2.0 Table 29-, Event Request Message Event Data Field Contents.

AlexanderAmelkin: @Journey-x, please pay more attention to the spec. Table 29-6 "Event Request Message Event Data Field Contents" specifies field values for "threshold", "discrete", and "OEM" sensors, not for the "Version Change" sensor type, values of Event Data 2 for which are covered by table 42-3 "Sensor Type Codes" and contain no `0xff`.

PR #328, opened 2022-03-02
fixes #172, Add support for editing FRU 'board mfg date'

doublechiang:
based on discussion https://github.com/ipmitool/ipmitool/pull/172#issuecomment-1031385132

### DIFF
--- a/include/ipmitool/ipmi_fru.h
+++ b/include/ipmitool/ipmi_fru.h
@@ -598,8 +598,10 @@ struct fru_picmgext_amc_link_desc_record {
 /* FRU Board manufacturing date */
 #define FRU_BOARD_DATE_UNSPEC 0 /* IPMI FRU Information Storage Definition
                                    v1.0 rev 1.3, Table 11-1 */
+
+static inline uint32_t SECS_FROM_1970_1996=820454400;
+
 static inline time_t ipmi_fru2time_t(void *mfg_date) {
-	const uint64_t secs_from_1970_1996 = 820454400;
 	uint32_t fru_ts = ipmi24toh(mfg_date);
 	time_t ts;
 
@@ -607,10 +609,21 @@ static inline time_t ipmi_fru2time_t(void *mfg_date) {
 		ts = IPMI_TIME_UNSPECIFIED;
 	}
 	else {
-		ts = fru_ts * 60 + secs_from_1970_1996;
+		ts = fru_ts * 60 + SECS_FROM_1970_1996;
 	}
 
 	return ts;
+}
+
+/**
+ * @brief Convert a host time_t from 1/1/1970 to IPMI mfg date
+ * @param[out]	mfg_date	at least 3 bytes buffer
+ * @returns the number of minutes from 1/1/1996 00:00
+ */
+static inline ipmi_time2fru(time_t sec, void *mfg_date) {
+	sec -= SECS_FROM_1970_1996;
+	sec/=60;
+	htoipmi24(sec, mfg_date);
 }
 
 typedef struct ipmi_fru_bloc {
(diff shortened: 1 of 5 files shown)


include/ipmitool/ipmi_fru.h
```diff
@@ -598,21 +598,34 @@ struct fru_picmgext_amc_link_desc_record {
 /* FRU Board manufacturing date */
 #define FRU_BOARD_DATE_UNSPEC 0 /* IPMI FRU Information Storage Definition
                                    v1.0 rev 1.3, Table 11-1 */
+
+static inline uint32_t SECS_FROM_1970_1996=820454400;
```
AlexanderAmelkin, line 602: Why did you demote it to `uint32_t` while moving it to global namespace?
Actually, I think it would be the best to use `time_t`. And it has to be `const`, on the contrary, `inline` doesn't seem to do anything here.

lib/ipmi_fru.c
```diff
@@ -4797,6 +4797,32 @@ f_type, uint8_t f_index, char *f_string)
...
+	if (f_type == 'b' && f_index == 9)		// use index '9' for editing 'board mfg date'
+	{
+		time_t total_secs;
+		uint32_t *board_mfg_date=&fru_data[3];
+
+		total_secs = ipmi_strptime("%Y-%m-%d %H:%M:%S", f_string);
+		ipmi_time2fru(total_secs, board_mfg_date);
+
+		checksum = 0;
+	/* Calculate Header Checksum */
```
AlexanderAmelkin, line 4809: Indent the comment to the same level with the code, please.

lib/ipmi_fru.c
```diff
@@ -4797,6 +4797,32 @@ f_type, uint8_t f_index, char *f_string)
 	/* Convert index from character to decimal */
 	f_index= f_index - 0x30;
 
+	if (f_type == 'b' && f_index == 9)		// use index '9' for editing 'board mfg date'
+	{
+		time_t total_secs;
+		uint32_t *board_mfg_date=&fru_data[3];
+
+		total_secs = ipmi_strptime("%Y-%m-%d %H:%M:%S", f_string);
+		ipmi_time2fru(total_secs, board_mfg_date);
```
AlexanderAmelkin, line 4806: looks like a range check is required here. we don't want values that exceed 24 bits, but `ipmi_strptime` is capable of returning greater values.

lib/ipmi_fru.c
```diff
@@ -4797,6 +4797,32 @@ f_type, uint8_t f_index, char *f_string)
...
+
+		checksum = 0;
+	/* Calculate Header Checksum */
+		for (i = 0; i < fru_section_len - 1; i++)
+		{
+			checksum += fru_data[i];
+		}
+		checksum = (~checksum) + 1;
+		fru_data[fru_section_len - 1] = checksum;
```
AlexanderAmelkin, line 4815: use `ipmi_csum()` please

AlexanderAmelkin: Also, it looks to me like all this FRU update code forgets to update the overall FRU checksum after modifying an area.

AlexanderAmelkin (changes requested):

lib/ipmi_fru.c
doublechiang: How should we check the range of the total_secs? 
The total_secs will be subtracted by SECS_FROM_1970_1996 and divided by 60 to get the minutes and mask down to three bytes. I think this is safe here.

lib/ipmi_fru.c
doublechiang: since we update the mfg date only and we do not change any length of the records. 
I don't think the common header field checksum will be changed.

doublechiang:
Please let me know if there are any changes required.

PR #359, opened 2022-08-30
sdr type sub command doesn't work when user specifiy hex value

### LINKED ISSUE #322: sdr type sub-command doesn't work when user specifiying hex value directly.
https://github.com/ipmitool/ipmitool/blob/39ca56bf33975b8a8b7e87928d67dc66366161da/lib/ipmi_sdr.c#L4584

the code tries to compare a user specified hex value against "0x".  eg, if user specify "0x80", the code tries to do strcmp("0x80", "0x"), and expect it to return 0. However, it seems strcmp() will return 0, if and only if two strings are exactly same. For this particular usage, I suggest to change the code to strncmp(type, "0x", 2); to force it to only compare first two characters.

coklm:
this fixes #322，please review

### DIFF
--- a/lib/ipmi_sdr.c
+++ b/lib/ipmi_sdr.c
@@ -4581,7 +4581,7 @@ ipmi_sdr_print_type(struct ipmi_intf *intf, char *type)
 		return 0;
 	}
 
-	if (!strcmp(type, "0x")) {
+	if (strncmp(type, "0x", 2)) {
 		/* begins with 0x so let it be entered as raw hex value */
 		if (str2uchar(type, &sensor_type) != 0) {
 			lprintf(LOG_ERR,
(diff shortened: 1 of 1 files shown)


AlexanderAmelkin:
[sdr type sub command doesn't work when user specifiy hex value]

must be:

```
sdr type: Fix use with type as hex value

This is a refix of 956ae2b372b612b4cf133ecb424f58b7edc418cb.

That commit changed `strncmp()` for `strcmp()` and thus broke
the check for hex value prefix in `sdr type` subcommand.

Resolves ipmitool/ipmitool#322

Signed-off-by: Your Name <your@e.mail>
```

[fix ipmitool sdr type "FRU State" will not found]

must be:

```
sdr type: Fix finding 'FRU State' type

The search for SDR type was using a wrong range.

Related to ipmitoo/ipmitool#322

Signed-off-by: Your Name <your@e.mail>
```

AlexanderAmelkin (changes requested):
Please fix the commit messages

PR #360, opened 2022-09-02
ipmi_fru.c: Provide missing function declarations

kraj:
Fixes build with clang-15+

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/lib/ipmi_fru.c
+++ b/lib/ipmi_fru.c
@@ -60,6 +60,13 @@ static const char *section_id[4] = {
 	"Board Section",
 	"Product Section"
 };
+/* From lib/ipmi_hpmfwupg.c: */
+uint16_t
+ipmi_intf_get_max_request_data_size(struct ipmi_intf * intf);
+
+/* From src/plugins/ipmi_intf.c: */
+uint16_t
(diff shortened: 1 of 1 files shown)


lib/ipmi_fru.c
```diff
@@ -60,6 +60,13 @@ static const char *section_id[4] = {
 	"Board Section",
 	"Product Section"
 };
+/* From lib/ipmi_hpmfwupg.c: */
+uint16_t
+ipmi_intf_get_max_request_data_size(struct ipmi_intf * intf);
+
+/* From src/plugins/ipmi_intf.c: */
+uint16_t
+ipmi_intf_get_max_response_data_size(struct ipmi_intf * intf);
```
AlexanderAmelkin, line 69: These must go to the respective header files, not here

AlexanderAmelkin (changes requested):

lib/ipmi_fru.c
kraj: in latest sources I see these functions are already defined in `include/ipmitool/ipmi_intf.h` after https://github.com/ipmitool/ipmitool/commit/1724b031 which fixes the problem at hand, so this patch is not needed anymore.

PR #371, opened 2022-10-05
Add bind_intf argument to specify network interface for LAN commands

dvanallen:
Optionally sets the SO_BINDTODEVICE socket option for LAN connections, to allow binding to a specific interface specified on the command line.  This has been tested successfully on bare-metal Linux servers.


AlexanderAmelkin:
Could you please explain why would anyone need that? Are the Linux routing rules not enough to establish the outbound interface? Why?

dvanallen:
Hi Alexander,

This is helpful when using a non-default VRF, because we can pass the VRF interface name to this new option, and then the Linux routing rules *for that specific VRF* will be used.

This is similar to the support provided by `ip vrf exec`, except that only works in environments with cgroups v2, and we'd like to support environments with only cgroups v1 support.

Some supporting docs are at https://docs.kernel.org/networking/vrf.html, which also notes "Processes can be “VRF aware” by binding a socket to the VRF device."

Happy to take feedback on this change!  I saw there are a few other places in the README / man page where the options are described, so I can work on getting those updated as part of this PR, if this is a feature you'd be open to accepting.

Thanks for taking a look.

AlexanderAmelkin:
It would be nice to add that explanation (maybe briefer version) to the commit log as well.

lib/ipmi_main.c
```diff
@@ -371,6 +373,17 @@ ipmi_main(int argc, char ** argv,
 	while ((argflag = getopt(argc, (char **)argv, OPTION_STRING)) != -1)
 	{
 		switch (argflag) {
+		case 'i':
+			if (bind_intf) {
+				free(bind_intf);
+				bind_intf = NULL;
```
AlexanderAmelkin, line 379: there is `free_n()` function for that in `include/ipmitool/helper.h`, please use it

AlexanderAmelkin (changes requested):

AlexanderAmelkin:
Also, it appears like your change is not compatible with macos and windows and must be conditionally disabled there.

src/plugins/ipmi_intf.c
```diff
@@ -236,6 +236,17 @@ ipmi_intf_session_set_hostname(struct ipmi_intf * intf, char * hostname)
 	intf->ssn_params.hostname = strdup(hostname);
 }
 
+void ipmi_intf_session_set_bind_intf(struct ipmi_intf *intf, char *bind_intf) {
+	if (intf->ssn_params.bind_intf) {
+		free(intf->ssn_params.bind_intf);
+		intf->ssn_params.bind_intf = NULL;
```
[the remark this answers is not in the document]
dvanallen: Done.

dvanallen:
Force-pushed to get the commit description updated.  Added conditional disablement and docs for the new option.  Could you please kick off the test workflows again?

src/plugins/ipmi_intf.c
```diff
@@ -399,6 +409,21 @@ ipmi_intf_socket_connect(struct ipmi_intf * intf)
...
+			#ifdef SO_BINDTODEVICE
+			int err = setsockopt(intf->fd, SOL_SOCKET, SO_BINDTODEVICE,
+					params->bind_intf, strlen(params->bind_intf) + 1);
+			if (err) {
+				lprintf(LOG_ERR, "Binding to network interface %s failed: %d",
+						params->bind_intf, err);
+				continue;
+			}
+			#else
+			lprintf(LOG_ERR, "Binding network interface is not supported on this platform.");
```
AlexanderAmelkin, line 422: I'd prefer the whole scope of changes related to this option to be left out if it is not supported, rather than listing the option and then printing that it is not supported.

src/plugins/ipmi_intf.c
```diff
@@ -399,6 +409,21 @@ ipmi_intf_socket_connect(struct ipmi_intf * intf)
 			continue;
 		}
 
+		if (params->bind_intf != NULL) {
+			#ifdef SO_BINDTODEVICE
```
AlexanderAmelkin, line 413: first, please place the `#ifdef` at the leftmost column.

second, if `SO_BINDTODEVICE` will ever change from a macro to an `enum` or `const`, this check will break. It's better to go the conventional feature-checking way, alter `configure.ac` with some extra `AC_TRY_COMPILE` to check if the code using `SO_BINDTODEVICE` can be compiled on the platform, then `AC_DEFINE()` some `HAS_BINDTODEVICE` macro, and then rely on that macro here.

Additional benefit to that approach is that you can check that autoconf definition in `ipmitool.1.in` as well and don't list in the `man` the option that is not supported on the platform.

AlexanderAmelkin (changes requested):

src/plugins/ipmi_intf.c
dvanallen: Done.

src/plugins/ipmi_intf.c
dvanallen: Thanks for the guidance here, and apologies for the delay in getting back.

I added the suggested `AC_TRY_COMPILE` macro, and used that throughout the source files, but I haven't found a good way to do a conditional change of the man page.  I considered adding an output variable based on the same `AC_TRY_COMPILE` check, which would be empty or hold the option help. That doesn't seem to work for multi-line strings, and would move the option help string all the way into configure.ac, which seems odd.

I note that we have an existing similar case with the dual bridge support for the `open` interface, where we check for feature support with an AC_TRY_COMPILE at configure time, but we always include the related options in the man page.

Would it be OK to tackle the platform-specific man page request in another PR (or do you have some more advice on how I should tackle this here)?

PR #378, opened 2022-11-16
fix-unalign-pointer-in-arm64

SliderLu:
For computers using the arm64 of the Apple chip, the link requires strict alignment of pointers in 32-bit form. In the structure vFlashstr, an 8-bit UINT-sized val is aligned with a (char *) str, resulting in a link error. It is necessary to adjust the consignment of val to 32bit UINT to solve this bug

Resolves ipmitool/ipmitoo#332
Signed-off-by: SliderLu <314238828@qq.com>


include/ipmitool/ipmi_delloem.h
```diff
@@ -344,7 +344,7 @@ typedef struct _power_headroom
 } __attribute__ ((packed)) POWER_HEADROOM;
 
 struct vFlashstr {
-	uint8_t val;
+	uint32_t val;
```
AlexanderAmelkin, line 347: It's actually unclear why a custom `struct vFlashstr` is used instead of generic `struct valstr` from `helper.h`.
The function `get_vFlash_compcode_str()` can as well be ditched and the generic `val2str()` can be used instead.
I think it's a good moment to get rid of that de-facto duplicate code.

SliderLu: I agree. I will commit it later

SliderLu:
add a commit of the issue

AlexanderAmelkin:
Thanks. Now please squash the commits together and fix the commit message according to the rules:
https://github.com/ipmitool/ipmitool/wiki/How-to-contribute#commit-preparation-rules

SliderLu:
Squashed and fixed the commit message.

AlexanderAmelkin:
The commit message still doesn't follow the 50/72 rule. Please fix.

SliderLu:
> The commit message still doesn't follow the 50/72 rule. Please fix.

Fiexed

AlexanderAmelkin (approved):

AlexanderAmelkin (approved):

PR #382, opened 2022-12-09
Do not require the IANA PEN registry file

vfazio:
Previously, ipmitool would fail to run if the local copy of the IANA PEN registry could not be parsed.

The registry is only a map and a missing file is already handled with the struct `ipmi_oem_info_dummy`. When the registry is not available the manufacturer will be "Unknown" but ipmitool will otherwise function.

Now, ipmitool will not consider a missing registry as fatal.

### DIFF
--- a/lib/ipmi_main.c
+++ b/lib/ipmi_main.c
@@ -853,10 +853,7 @@ ipmi_main(int argc, char ** argv,
 	}
 
 	/* load the IANA PEN registry */
-	if (ipmi_oem_info_init()) {
-		lprintf(LOG_ERR, "Failed to initialize the OEM info dictionary");
-		goto out_free;
-	}
+	ipmi_oem_info_init();
 
 	/* run OEM setup if found */
 	if (oemtype &&
--- a/lib/ipmi_strings.c
+++ b/lib/ipmi_strings.c
@@ -1719,39 +1719,30 @@ out:
 	return rc;
 }
 
-int ipmi_oem_info_init()
+void ipmi_oem_info_init()
 {
 	oem_valstr_list_t terminator = { { -1, NULL}, NULL }; /* Terminator */
 	oem_valstr_list_t *oemlist = &terminator;
 	bool free_strings = true;
-	size_t count;
-	int rc = -4;
+	int count;
 
 	lprintf(LOG_INFO, "Loading IANA PEN Registry...");
 
 	if (ipmi_oem_info) {
 		lprintf(LOG_INFO, "IANA PEN Registry is already loaded");
-		rc = 0;
 		goto out;
 	}
 
-	if (!(count = oem_info_list_load(&oemlist))) {
-		/*
-		 * We can't identify OEMs without a loaded registry.
-		 * Set the pointer to dummy and return.
-		 */
-		ipmi_oem_info = ipmi_oem_info_dummy;
-		goto out;
+	if ((count = oem_info_list_load(&oemlist)) < 1) {
+		lprintf(LOG_WARN, "Failed to load entries from IANA PEN Registry");
+		count = 0;
 	}
 
 	/* In the array was allocated, don't free the strings at cleanup */
 	free_strings = !oem_info_init_from_list(oemlist, count);
 
-	rc = IPMI_CC_OK;
-
 out:
 	oem_info_list_free(&oemlist, free_strings);
-	return rc;
 }
 
 void ipmi_oem_info_free()
--- a/include/ipmitool/ipmi_strings.h
+++ b/include/ipmitool/ipmi_strings.h
@@ -55,7 +55,7 @@ extern const struct valstr ipmi_integrity_algorithms[];
 extern const struct valstr ipmi_encryption_algorithms[];
 extern const struct valstr ipmi_user_enable_status_vals[];
 extern const struct valstr *ipmi_oem_info;
-int ipmi_oem_info_init();
+void ipmi_oem_info_init();
 void ipmi_oem_info_free();
 
 extern const struct valstr picmg_frucontrol_vals[];


vfazio:
With PEN registry:
```
# ipmitool -I serial-terminal -D /dev/ttyS2:115200 mc info
Device ID                 : 1
Device Revision           : 1
Firmware Revision         : 3.17
IPMI Version              : 2.0
Manufacturer ID           : 27317
Manufacturer Name         : Extreme Engineering Solutions, Inc
Product ID                : 4478 (0x117e)
Product Name              : Unknown (0x117E)
Device Available          : yes
Provides Device SDRs      : yes
Additional Device Support :
    Sensor Device
    SEL Device
    FRU Inventory Device
    IPMB Event Generator

```
Without:

```
# rm /usr/share/misc/enterprise-numbers 
# ipmitool -I serial-terminal -D /dev/ttyS2:115200 mc info
IANA PEN registry open failed: No such file or directory
Device ID                 : 1
Device Revision           : 1
Firmware Revision         : 3.17
IPMI Version              : 2.0
Manufacturer ID           : 27317
Manufacturer Name         : Unknown (0x6AB5)
Product ID                : 4478 (0x117e)
Product Name              : Unknown (0x117E)
Device Available          : yes
Provides Device SDRs      : yes
Additional Device Support :
    Sensor Device
    SEL Device
    FRU Inventory Device
    IPMB Event Generator

```

lib/ipmi_strings.c
```diff
@@ -1735,7 +1735,7 @@ int ipmi_oem_info_init()
 		goto out;
 	}
 
-	if (!(count = oem_info_list_load(&oemlist))) {
+	if ((count = oem_info_list_load(&oemlist)) < 1) {
```
AlexanderAmelkin, line 1738: As I look at the contents of this conditional block now, I suspect it's actually not correct to set `ipmi_oem_info` to `ipmi_oem_dummy` in any case. Instead, `oem_info_init_from_list()` should be taught to properly build an info list (with head and tail) from an empty input.

lib/ipmi_strings.c
```diff
@@ -1724,7 +1724,7 @@ int ipmi_oem_info_init()
 	oem_valstr_list_t terminator = { { -1, NULL}, NULL }; /* Terminator */
 	oem_valstr_list_t *oemlist = &terminator;
 	bool free_strings = true;
-	size_t count;
+	int count;
```
AlexanderAmelkin, line 1727: Although I don't agree with the solution in general, this is a good catch anyway as `oem_info_list_load()` may return a negative value.

lib/ipmi_main.c
```diff
@@ -854,8 +854,7 @@ ipmi_main(int argc, char ** argv,
 
 	/* load the IANA PEN registry */
 	if (ipmi_oem_info_init()) {
-		lprintf(LOG_ERR, "Failed to initialize the OEM info dictionary");
-		goto out_free;
+		lprintf(LOG_WARN, "Failed to initialize the OEM info dictionary");
```
AlexanderAmelkin, line 857: Looks like the function may not fail now and can be made `void`.
The warning should be moved into that function instead.

AlexanderAmelkin (changes requested):

lib/ipmi_strings.c
vfazio: So you want to drop the `ipmi_oem_dummy` struct completely? If so, i'll work that into the reworked commits

vfazio: This has implications for `oem_info_init_from_list` because if it can't allocate memory, it uses the dummy list. If we get rid of it, then ipmitool will need to always be able to alloc at least `(head_entries + tail_entries + 1) * sizeof(*ipmi_oem_info)` bytes of memory instead of being able to rely on that static struct.

If that memory allocation fails, then `ipmi_oem_info_init` needs to remain non-void to propagate the failure.

vfazio: I think i understood your proposal. should be addressed

vfazio:
I retested with these changes

```
# ipmitool  -I serial-terminal -D /dev/ttyS2:115200 mc info
Device ID                 : 1
Device Revision           : 1
Firmware Revision         : 3.17
IPMI Version              : 2.0
Manufacturer ID           : 27317
Manufacturer Name         : Extreme Engineering Solutions, Inc
Product ID                : 4479 (0x117f)
Product Name              : Unknown (0x117F)
Device Available          : yes
Provides Device SDRs      : yes
Additional Device Support :
    Sensor Device
    SEL Device
    FRU Inventory Device
    IPMB Event Generator


# rm /usr/share/misc/enterprise-numbers

# ipmitool -I serial-terminal -D /dev/ttyS2:115200 mc info
IANA PEN registry open failed: No such file or directory
Failed to load entries from IANA PEN Registry
Device ID                 : 1
Device Revision           : 1
Firmware Revision         : 3.17
IPMI Version              : 2.0
Manufacturer ID           : 27317
Manufacturer Name         : Unknown (0x6AB5)
Product ID                : 4479 (0x117f)
Product Name              : Unknown (0x117F)
Device Available          : yes
Provides Device SDRs      : yes
Additional Device Support :
    Sensor Device
    SEL Device
    FRU Inventory Device
    IPMB Event Generator

# touch /usr/share/misc/enterprise-numbers

# ipmitool -I serial-terminal -D /dev/ttyS2:115200 mc info 
Failed to load entries from IANA PEN Registry
Device ID                 : 1
Device Revision           : 1
Firmware Revision         : 3.17
IPMI Version              : 2.0
Manufacturer ID           : 27317
Manufacturer Name         : Unknown (0x6AB5)
Product ID                : 4479 (0x117f)
Product Name              : Unknown (0x117F)
Device Available          : yes
Provides Device SDRs      : yes
Additional Device Support :
    Sensor Device
    SEL Device
    FRU Inventory Device
    IPMB Event Generator


```

AlexanderAmelkin (approved):

PR #389, opened 2022-12-21
lanp: Fix error response from Unsupported Parameter lookup

### LINKED ISSUE #388: Unsupported LAN Parameter lookup command returns an error
When using `ipmitool lan print`, an error (exit status 1) is returned for an Unsupported Parameter in versions which include the change https://github.com/ipmitool/ipmitool/commit/351dad24a26f56580ba6028a267c4924b7faa4cb.

**IPMITOOL Version**

```none
ipmitool -V
ipmitool version 1.8.19.4.g63d59a5
```

**To Reproduce**
1. Using Dell R740 server with Linux 5.15
2. Run `ipmitool -vvvvv lan print`
3. Observe the abnormal log message and exit code 1:
    ```none
    Get LAN Parameter 'Bad Password Threshold' command failed: Unsupported parameter
    ```

**Expected behavior**
Function `get_lan_param_select` should not return NULL, as described in the function comment at https://github.com/ipmitool/ipmitool/blob/63d59a50a6809e415cd7d8915533f236f28865f3/lib/ipmi_lanp.c#L179-L181. The `ipmitool lan print` command should return exit code 0.

dvanallen:
Return a pointer to the lan_param instead of NULL in the case when the parameter is not supported.

Resolves ipmitool/ipmitool#388
Signed-off-by Daniel Van Allen <dvanallen@google.com>

### DIFF
--- a/lib/ipmi_lanp.c
+++ b/lib/ipmi_lanp.c
@@ -236,8 +236,12 @@ get_lan_param_select(struct ipmi_intf *intf, uint8_t chan, int param, int select
 		/* We treat them as valid but empty response */
 		p->data = NULL;
 		p->data_len = 0;
-		rc = p;
-		/* fall through */
+		lprintf(LOG_INFO, "Get LAN Parameter '%s' command unsupported: %s",
+			p->desc,
+			specific_val2str(rsp->ccode,
+			                 get_lan_cc_vals,
+			                 completion_code_vals));
+		return p;
 	default:
 		/* other completion codes are treated as error */
 		lprintf(LOG_INFO, "Get LAN Parameter '%s' command failed: %s",


osenft (dismissed):

osenft (approved):

osenft (dismissed):

osenft:
Could you please add a few lines about the broken behavior w/o the fix, the expected behavior and which platform you tested this on?

osenft:
> Could you please add a few lines about the broken behavior w/o the fix, the expected behavior and which platform you tested this on?

Oh, sorry, you have it in the issue #388. My bad.

dvanallen:
@AlexanderAmelkin please let me know if you'd like me to keep syncing this branch to master, I don't want to change it out from underneath you or make @osenft's review stale if you're actively trying to review it.

lib/ipmi_lanp.c
```diff
@@ -245,7 +245,7 @@ get_lan_param_select(struct ipmi_intf *intf, uint8_t chan, int param, int select
 			specific_val2str(rsp->ccode,
 			                 get_lan_cc_vals,
 			                 completion_code_vals));
-		return NULL;
+		return rc;
```
AlexanderAmelkin, line 248: In the commit log you claim that you return the pointer "in the case when the parameter is not supported", but it's not actually true as the `default` case covers all possible non-successful completion codes.
Moreover, the 'parameter is not supported' case is handled above in line 233.
Looks more like the above cases must not fall through to `default` and must print their own message, slightly different from what the `default` case must print.

AlexanderAmelkin (changes requested):

lib/ipmi_lanp.c
dvanallen: Note that in the default case `rc` will be NULL, and that we return `rc` as a way of returning NULL several times earlier in the function.

When looking at this issue, I realized that we currently set `rc` from this `case 0xcc` block but then we never use it, so I was thinking that my change here must have been the intended change that got missed somehow.

I do agree that's a bit tricky with the fall through, so I updated this case to add a different log message and return the pointer.

osenft: Alexander, I think it's worth pointing out that the specific case in question (parameter not supported) was working fine up to commit 351dad24a26f56580ba6028a267c4924b7faa4cb. At this point the specific log message was removed and the fall-through to the default case was added. I think it was at that point where the code change was actually incomplete as the "rc = p" was added, while the default case has "return NULL" instead of "return p".

I believe that Daniel's original change fixed this oversight by simply changing the default to "return rc". 

In the latest version of the code "rc" is always NULL and cases of "return rc" should be replaced with "return NULL".

Can you please clarify whether 351dad24a26f56580ba6028a267c4924b7faa4cb changing the log message was intentional? If so, Daniel's latest change effectively undoes that.

I believe the commit log saying "you return the pointer "in the case when the parameter is not supported"" was correct, since in the default case rc = NULL, so in the default case nothing changed.

rustydb:
Any updates on the outlook for this?

PR #393, opened 2023-01-12
configure.ac: allow disabling registry downloads

vfazio:
Some environments require reproducible builds. Since the IANA PEN registry is constantly updating and there is no snapshot available, installing ipmitool via `make install` is not reproducible.

Provide a configure mechanism to disable the registry download/install..


vfazio:
As part of this change, the registry download/install can be skipped via:

```
./configure --disable-registry-download
```

This causes the checks for wget/curl to get skipped and ensures DOWNLOAD is substituted as FALSE. 

I restructured the conditionals but still gave precedence to wget.

AlexanderAmelkin (approved):

